PR #1, opened 2015-07-30
fix more windows warnings

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_windows/85/warnings34Result/


dirk-thomas:
The fix for Windows has been rolled into #2.

PR #2, opened 2015-07-30
add subscriber to flip image

dirk-thomas:


esteve:
+1, please squash the commits before merging

dirk-thomas:
Since both commits change different things which are not related to each other I would rather keep them as separate commits.

esteve:
@dirk-thomas no problem.

wjwwood:
+1

PR #3, opened 2015-07-30
Showimage

esteve:
Adds an executable to visualize images coming from a subscription.

Depends on #2 

Connects to ros2/ros2#54


esteve:
@dirk-thomas @jacquelinekay @tfoote @wjwwood ping

cam2image/src/cam2image.cpp
```diff
@@ -99,6 +99,7 @@ int main(int argc, char * argv[])
         convert_frame_to_message(flipped_frame, i, msg);
       }
       std::cout << "Publishing image #" << i << std::endl;
+      msg->header.frame_id = std::to_string(i);
```
dirk-thomas: Why is this added here? That is already done in https://github.com/ros2/demos/blob/4f409e9e89796143a94b9fa1cd18988d64edb64f/cam2image/src/cam2image.cpp#L52

cam2image/src/showimage.cpp
```diff
@@ -0,0 +1,66 @@
...
+  } else if(encoding == "rgb8") {
+    return CV_8UC3;
+  } else if(encoding == "mono16") {
+    return CV_16SC1;
+  } else if(encoding == "rgba8") {
+    return CV_8UC4;
+  } else {
+    throw std::runtime_error("Unsupported encoding type");
+  }
+}
```
dirk-thomas: The same code already exists in the other cpp file. I would suggest moving it to a header and reusing it in both.

cam2image/src/showimage.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+void showImage(const sensor_interfaces::msg::Image::ConstSharedPtr & msg) {
+  std::cout << "Received image #" << msg->header.frame_id << std::endl;
```
dirk-thomas: `printf` instead.

cam2image/src/showimage.cpp
dirk-thomas: Never mind - this is the opposite direction.

cam2image/src/showimage.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  cv::imshow("view", frame);
+  cv::waitKey(1);
+}
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+
+  auto node = rclcpp::node::Node::make_shared("cam2image");
```
dirk-thomas: This should have a different node name.

cam2image/src/showimage.cpp
```diff
@@ -0,0 +1,66 @@
...
+  } else if(encoding == "mono16") {
+    return CV_16SC1;
+  } else if(encoding == "rgba8") {
+    return CV_8UC4;
+  } else {
+    throw std::runtime_error("Unsupported encoding type");
+  }
+}
+
+void showImage(const sensor_interfaces::msg::Image::ConstSharedPtr & msg) {
```
dirk-thomas: Should this be lower case underscore instead?

dirk-thomas:
The node for showing images is a bit weird in the `cam2image` package. Should we either split it into a separate packages or rename this package to describe better what it will contain?

esteve:
@dirk-thomas I'd rename the package, what would be a suitable name? `imagetools` ?

dirk-thomas:
`image_tools` sounds fine to me.

cam2image/src/cam2image.cpp
esteve: Removed.

cam2image/src/showimage.cpp
esteve: Used a `cout` and a `stringstream` to avoid interleaving.

cam2image/src/showimage.cpp
esteve: Fixed.

cam2image/src/showimage.cpp
esteve: Done.

esteve:
@dirk-thomas I just renamed the package to `image_tools`

cam2image/src/showimage.cpp
dirk-thomas: This could be a single line with `printf` instead of three.

image_tools/package.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <package format="2">
-  <name>cam2image</name>
+  <name>image_tools</name>
   <version>0.0.0</version>
   <description>A node to read images form a camera and publish them on a topic.</description>
```
tfoote: The description could use a slight extension for the new scope.

tfoote:
One comment on the description but otherwise lgtm

image_tools/package.xml
esteve: Done.

wjwwood:
lgtm

image_tools/package.xml
dirk-thomas: This comments does not seem to be addressed in the merged patch yet.

esteve: I committed it to the qos branch by mistake.

esteve: Fixed in master.

image_tools/src/showimage.cpp
```diff
@@ -0,0 +1,68 @@
...
+    return CV_16SC1;
+  } else if(encoding == "rgba8") {
+    return CV_8UC4;
+  } else {
+    throw std::runtime_error("Unsupported encoding type");
+  }
+}
+
+void show_image(const sensor_interfaces::msg::Image::ConstSharedPtr & msg) {
+  std::stringsteam ss;
```
dirk-thomas: This change breaks the build. Please fix the syntax.

jacquelinekay: I committed the fix for this syntax error to master in b603bfe739e2021ec7b64a0f34c349e02a57377d.

esteve: @jacquelinekay thanks! I pushed a change to include `#include <sstream>` right at the same time.

dirk-thomas:
@esteve With the windows build being back up again please revisit the new warnings introduced by this PR: http://ci.ros2.org/job/ros2_batch_ci_windows/133/warnings34Result/category.63481637/package.-2136935613/

esteve:
@dirk-thomas I committed this, but haven't merged it to master:

https://github.com/ros2/demos/commit/f3ab25dbe4cd928d1b0073442a7bbf67f2e9f82a

this removes the warning you linked to, though it doesn't fix the existing warnings.

dirk-thomas:
Looking at the surrounding code it looks like that the allocated `data` is leaked for every frame. Before continuing on the iterator the first thing should be to fix the memory leak. May be loaning the ROS image data directly to OpenCV will avoid the whole copy?

esteve:
@dirk-thomas I changed it to use the `data()` pointer from the `std::vector`:

https://github.com/ros2/demos/commit/38e89cd6b4c8e01284d33ba4eea11c66b8138721

I'll start a Windows CI job when Jose is done with the Windows slave.

PR #4, opened 2015-07-30
Added support for QoS profiles

esteve:
Connects to ros2/ros2#66

### DIFF
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -58,7 +58,11 @@ int main(int argc, char * argv[])
 
   auto node = rclcpp::node::Node::make_shared("cam2image");
 
-  auto pub = node->create_publisher<sensor_interfaces::msg::Image>("image", 10);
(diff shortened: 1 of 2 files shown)


dirk-thomas:
+1

image_tools/src/cam2image.cpp
```diff
@@ -58,7 +58,11 @@ int main(int argc, char * argv[])
 
   auto node = rclcpp::node::Node::make_shared("cam2image");
 
-  auto pub = node->create_publisher<sensor_interfaces::msg::Image>("image", 10);
+  rmw_qos_policies_t custom_qos_policies = rmw_qos_policies_default;
+  custom_qos_policies.depth = 10;
```
jacquelinekay: Will this number be configurable at some point (such as via command line args or a GUI tool)?

esteve: It should be, but I didn't want to make more changes than necessary, just update the code for the QoS changes.

jacquelinekay:
+1

PR #5, opened 2015-08-04
Demo for intra process communication

wjwwood:
Work in progress, not ready for review.

Connects to ros2/ros2#59

### DIFF
--- a/intra_process_comms/CMakeLists.txt
+++ b/intra_process_comms/CMakeLists.txt
@@ -1,0 +1,24 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(intra_process_comms_example)
+
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rmw REQUIRED)
+find_package(rmw_implementation REQUIRED)
+find_package(std_msgs REQUIRED)
+
+ament_package()
+
+##
+## Example of intra-process comms
+##
+
+add_executable_for_each_rmw_implementations(intrapc
+  src/intrapc.cpp
+  TARGET_DEPENDENCIES "rclcpp" "std_msgs"
+  INSTALL)
--- a/intra_process_comms/package.xml
+++ b/intra_process_comms/package.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>intra_process_comms_example</name>
+  <version>0.0.0</version>
+  <description>Prototype of intra-process communications.</description>
+  <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <build_depend>rclcpp</build_depend>
+  <build_depend>rmw_implementation</build_depend>
+  <build_depend>std_msgs</build_depend>
+
+  <exec_depend>rclcpp</exec_depend>
+  <exec_depend>rmw_implementation</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
+
+  <export><build_type>ament_cmake</build_type></export>
+</package>
--- a/intra_process_comms/src/intrapc.cpp
+++ b/intra_process_comms/src/intrapc.cpp
@@ -1,0 +1,71 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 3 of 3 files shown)


intra_process_comms/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rmw REQUIRED)
+find_package(rmw_implementation REQUIRED)
+find_package(std_msgs REQUIRED)
+
+ament_package()
+
+# This file provides the add_executable_for_each_rmw_impl function.
+include(cmake/add_executable_for_each_rmw_impl.cmake)
```
dirk-thomas: Please use ros2/rmw_implementation#3 instead.

intra_process_comms/package.xml
```diff
@@ -0,0 +1,23 @@
...
+<package format="2">
+  <name>intra_process_comms_example</name>
+  <version>0.0.0</version>
+  <description>Prototype of intra-process communications.</description>
+  <maintainer email="william@osrfoundation.org">William Woodall</maintainer>
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <buildtool_depend>rosidl_default_generators</buildtool_depend>
```
dirk-thomas: The package does not contain any messages and therefore neither requires `rosidl_default_generators` nor `rosidl_default_runtime`.

intra_process_comms/src/intrapc.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <functional>
+#include <iostream>
+
+#include <rclcpp/rclcpp.hpp>
+
+#include <std_msgs/msg/string.hpp>
+
+void on_message(const std_msgs::msg::String::ConstSharedPtr & msg)
+{
+  std::cout << "I heard [" << msg->data << "]" << std::endl;
```
dirk-thomas: Please use `printf`.

Same below.

intra_process_comms/src/intrapc.cpp
```diff
@@ -0,0 +1,72 @@
...
+  rclcpp::sleep_for(1_s);
+  if (msg->data.at(msg->data.length() - 1) == '9')
+  {
```
dirk-thomas: Style.

intra_process_comms/src/intrapc.cpp
```diff
@@ -0,0 +1,72 @@
...
+    std::cerr << "Invalid publisher!" << std::endl;
+  }
+}
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+
+  auto node1 = rclcpp::Node::make_shared("node1");
+  auto node2 = rclcpp::Node::make_shared("node2");
```
dirk-thomas: Don't they have to enable intra process comm here (like https://github.com/ros2/system_tests/pull/28/files#diff-fe711788916f61638c377d72e77584a5R28)?

intra_process_comms/src/intrapc.cpp
```diff
@@ -0,0 +1,72 @@
...
+// limitations under the License.
+
+#include <functional>
+#include <iostream>
+
+#include <rclcpp/rclcpp.hpp>
+
+#include <std_msgs/msg/string.hpp>
+
+void on_message(const std_msgs::msg::String::ConstSharedPtr & msg)
```
dirk-thomas: The reference `&` is not needed for shared pointers.

Same below.

jacquelinekay:
The description says it's not ready for review, but I assume this is outdated because you've moved this set of changes into "In Review"...

intra_process_comms/src/intrapc.cpp
jacquelinekay: Do you plan on adding output that shows intra-process communication is being used (like comparing the address of the message pointer)?

wjwwood: > The reference & is not needed for shared pointers.

This is the signature for subscription callbacks, e.g.: https://github.com/ros2/examples/blob/master/rclcpp_examples/src/topics/listener.cpp#L22

Why not take them by reference?

dirk-thomas: A reference is basically a pointer. If the caller lets the shared pointer go out of scope the reference this function is holding becomes invalid. Therefore I think the function should receive its own shared pointer which will then also contribute to the reference counting.

wjwwood: Part of the contract of the subscription callback is that this will not happen. I'm not going to change this here until we change it in all of our callbacks.

wjwwood: I changed the one I control here: cb42509

wjwwood:
Yeah, this one is still pretty basic. I think it's worth reviewing now and merging and it will get expanded and updated when I write the demo tutorial.

I addressed the comments in the latest commits.

intra_process_comms/src/intrapc.cpp
dirk-thomas: Sure, I have created ros2/ros2#94 to keep track of fixing it in the existing code.

esteve:
+1

PR #7, opened 2015-08-12
Minimal implementation of realtime control demo

jacquelinekay:
This pull request adds:

* `pendulum_control` package
* `pendulum_msgs` package for custom message types. (Originally I considered using `sensor_msgs::JointState`, but since it contains a string field, it is not a fixed size message.)
* `RttExecutor`: an executor instrumented for collecting data for real-time performance statistics
* pendulum_demo.cpp: provides a main function that intializes rclcpp, nodes, publishers/subscribers, and executor. Print output to screen in a separate, low-priority thread.
* `PendulumController`: Store PID properties, provide the next command message based on the received sensor message.
* `PendulumMotor`: Abstraction for the physical state of the pendulum system. Change the state based on the command message and provide the next sensor message. Runs a separate "physics update" thread to update the state of the pendulum.

The `README.md` file contains more information about how to configure your system to see "soft" real-time performance results.

### DIFF
--- a/pendulum_control/CMakeLists.txt
+++ b/pendulum_control/CMakeLists.txt
@@ -1,0 +1,38 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(pendulum_control)
+
+if(APPLE OR WIN32)
+  message(STATUS "The pendulum_control demo is only supported on Linux. Package will not be built.")
+  return()
+endif()
+
+find_package(ament_cmake REQUIRED)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+
+find_package(rclcpp REQUIRED)
+find_package(rmw_implementation REQUIRED)
+find_package(pendulum_msgs REQUIRED)
+find_package(rttest)
+
+if(NOT rttest_FOUND)
+  message(STATUS "rttest not found. pendulum_control package will not be built.")
+  return()
+endif()
+
+add_executable_for_each_rmw_implementations(pendulum_demo
+  src/pendulum_demo.cpp
+  TARGET_DEPENDENCIES
+  "rclcpp"
+  "pendulum_msgs"
+  "rttest"
+  INSTALL
+)
+
+if(AMENT_ENABLE_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+endif()
+
+ament_package()
--- a/pendulum_control/src/pendulum_controller.hpp
+++ b/pendulum_control/src/pendulum_controller.hpp
@@ -1,0 +1,136 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef PENDULUM_DEMO_PENDULUM_CONTROLLER_HPP_
+#define PENDULUM_DEMO_PENDULUM_CONTROLLER_HPP_
+
+#include <chrono>
+
+#include <pendulum_msgs/msg/joint_command.hpp>
+#include <pendulum_msgs/msg/joint_state.hpp>
+
+#ifndef PI
+#define PI 3.14159265359
+#endif
+
+namespace pendulum_control
+{
+
+struct PIDProperties
+{
+  // Properties of a PID controller
+  double p = 1;
+  double i = 0;
+  double d = 0;
+  double command = PI / 2;
+};
+
+class PendulumController
+{
+public:
+  PendulumController(std::chrono::nanoseconds period, PIDProperties pid)
+  : publish_period_(period), pid_(pid),
+    command_message_(std::make_shared<pendulum_msgs::msg::JointCommand>()),
+    message_ready_(false)
+  {
+    command_message_->position = pid_.command;
+    dt_ = publish_period_.count() / (1000.0 * 1000.0 * 1000.0);
+    if (isnan(dt_) || dt_ == 0) {
+      throw std::runtime_error("Invalid dt_ calculated in PendulumController constructor");
+    }
+  }
+
+  // Calculate new command based on new sensor state and PID controller properties
+  void on_sensor_message(const pendulum_msgs::msg::JointState::SharedPtr msg)
+  {
+    ++messages_received;
+
+    if (isnan(msg->position)) {
+      throw std::runtime_error("Sensor value was NaN in on_sensor_message callback");
+    }
+    double error = pid_.command - msg->position;
+    double p_gain = pid_.p * error;
+    i_gain_ = pid_.i * (i_gain_ + error * dt_);
+    double d_gain = pid_.d * (error - last_error_) / dt_;
+    last_error_ = error;
+
+    // TODO consider filtering the PID output
+    command_message_->position = msg->position + p_gain + i_gain_ + d_gain;
+    // limits
+    if (command_message_->position > PI) {
+      command_message_->position = PI;
+    } else if (command_message_->position < 0) {
+      command_message_->position = 0;
+    }
+
+    if (isnan(command_message_->position)) {
+      throw std::runtime_error("Resulting command was NaN in on_sensor_message callback");
+    }
+    message_ready_ = true;
+  }
+
+  const pendulum_msgs::msg::JointCommand::SharedPtr get_next_command_message()
+  {
+    return command_message_;
+  }
+
+  bool next_message_ready() const
+  {
+    return message_ready_;
+  }
+
+  std::chrono::nanoseconds get_publish_period() const
+  {
+    return publish_period_;
+  }
+
+  void set_pid_properties(const PIDProperties & properties)
+  {
+    pid_ = properties;
+  }
+
+  const PIDProperties & get_pid_properties() const
+  {
+    return pid_;
+  }
+
+  void set_command(double command)
+  {
+    pid_.command = command;
+  }
+
+  double get_command() const
+  {
+    return pid_.command;
+  }
+
+  // gather statistics
+  size_t messages_received = 0;
+
+private:
+  // controller should publish less frequently than the motor
+  std::chrono::nanoseconds publish_period_;
+  PIDProperties pid_;
+  pendulum_msgs::msg::JointCommand::SharedPtr command_message_;
+  bool message_ready_;
+
+  // state for PID controller
+  double last_error_ = 0;
+  double i_gain_ = 0;
+  double dt_;
+};
+
+}  /* namespace pendulum_demo */
+
+#endif  /* PENDULUM_DEMO_PENDULUM_CONTROLLER_HPP_ */
--- a/pendulum_control/src/pendulum_demo.cpp
+++ b/pendulum_control/src/pendulum_demo.cpp
@@ -1,0 +1,188 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <rclcpp/memory_strategies.hpp>
+#include <rclcpp/rclcpp.hpp>
+#include <rclcpp/strategies/message_pool_memory_strategy.hpp>
+
+#include <rttest/rttest.h>
+
+#include <pendulum_msgs/msg/joint_command.hpp>
+#include <pendulum_msgs/msg/joint_state.hpp>
+
+#include "pendulum_controller.hpp"
+#include "pendulum_motor.hpp"
+#include "rtt_executor.hpp"
+
+struct OutputThreadArgs
+{
+  std::shared_ptr<pendulum_control::RttExecutor> executor;
+  std::shared_ptr<pendulum_control::PendulumMotor> pendulum_motor;
+  std::shared_ptr<pendulum_control::PendulumController> pendulum_controller;
+
+  OutputThreadArgs(std::shared_ptr<pendulum_control::RttExecutor> exec,
+    std::shared_ptr<pendulum_control::PendulumMotor> motor,
+    std::shared_ptr<pendulum_control::PendulumController> controller)
+  : executor(exec), pendulum_motor(motor), pendulum_controller(controller) {}
+};
+
+void * live_output_thread(void * args)
+{
+  OutputThreadArgs * ptr_args = static_cast<OutputThreadArgs *>(args);
+  auto executor = ptr_args->executor;
+  auto pendulum_motor = ptr_args->pendulum_motor;
+  auto pendulum_controller = ptr_args->pendulum_controller;
+
+  // background task: don't want to interfere with performance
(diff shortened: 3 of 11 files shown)


pendulum_control/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+project(pendulum_control)
+
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rmw_implementation REQUIRED)
+find_package(sensor_msgs REQUIRED)
```
dirk-thomas: Not listed in the manifest.

pendulum_control/src/pendulum_motor.hpp
```diff
@@ -0,0 +1,189 @@
...
+      state_.velocity += state_.acceleration * dt_;
+      state_.position += state_.velocity * dt_;
+      state_.position = fmod(state_.position, 2 * PI);
+      assert(!isnan(state_.position));
+      sensor_message_->velocity = state_.velocity;
+      sensor_message_->position = state_.position;
+
+      message_ready_ = true;
+      // high resolution sleep
+      clock_nanosleep(CLOCK_REALTIME, 0, &physics_update_timespec_, NULL);
```
adolfo-rt: Use `CLOCK_MONOTONIC` instead, so physics aren't affected by system time changes (NTP slew, user calling `date --set=Fri May 4 13:38:56 CEST 1979`)

pendulum_control/src/pendulum_controller.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+#ifndef PENDULUM_DEMO_PENDULUM_CONTROLLER_HPP_
+#define PENDULUM_DEMO_PENDULUM_CONTROLLER_HPP_
+
+#include <chrono>
+
+#include <pendulum_msgs/msg/joint_command.hpp>
+#include <pendulum_msgs/msg/joint_state.hpp>
+
+#ifndef PI
```
adolfo-rt: Anecdotal note: I always used `M_PI` from `<cmath>`, but just realized that it's not part of the C or C++ standards, but an extension of most POSIX systems :-0

pendulum_msgs/msg/JointState.msg
```diff
@@ -0,0 +1,3 @@
+float64 position
```
adolfo-rt: It's fine to use this for demo purposes. What I'd like to see in the long run is the ability to preallocate messages in publishers and subscribers, so dynamically-sized fields can be used. Something like:
```c++
sensor_msgs::msg::JointState msg;
msg.name.push_back("pendulum");
msg.position.resize(1, 0.0);
msg.velocity.resize(1, 0.0);
msg.effort.resize(1, 0.0);

publisher.preallocate(msg);  // optional
subscriber.preallocate(msg); //
```
`preallocate` would initialize with `msg` all internal buffers in the non real-time path. When publishing and subscribing from the real-time path, no further allocations are required as long as the vector and string sizes remain the same.

I don't know if there are details on how DDS works that would add additional complexity to this argument.

pendulum_control/src/pendulum_motor.hpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    return state_;
+  }
+
+  void set_state(const PendulumState & state)
+  {
+    state_ = state;
+  }
+
+  PendulumProperties get_properties() const
```
dirk-thomas: Return non-primitive type as const reference.

pendulum_msgs/msg/JointState.msg
dirk-thomas: Since the message size is unknown based on the definition I don't think we can make the complete pipeline preallocated here. That is the reason why you can specify upper bounds now in the message definition. Only then we can ensure to preallocate data objects throughout the pipeline.

pendulum_control/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+      "rclcpp"
+      "${middleware_impl_tmp}"
+      "pendulum_msgs"
+      "rttest"
+    )
+
+    install(TARGETS ${executable}__${middleware_impl_tmp} DESTINATION bin)
+  endforeach()
+
+endfunction()
```
dirk-thomas: We should not copy this code block into every package but reuse the function from a known package. Any suggestions where this should go to? May be into `rmw`?

pendulum_control/src/pendulum_controller.hpp
```diff
@@ -0,0 +1,128 @@
...
+  {
+    return publish_period_;
+  }
+
+  void set_pid_properties(const PIDProperties & properties)
+  {
+    pid_ = properties;
+  }
+
+  PIDProperties get_pid_properties() const
```
dirk-thomas: Return non-primitive type as const reference.

pendulum_control/src/pendulum_controller.hpp
```diff
@@ -0,0 +1,128 @@
...
+  {
+    return pid_;
+  }
+
+  void set_command(const double command)
```
dirk-thomas: No const for non-reference arguments.

pendulum_control/src/pendulum_controller.hpp
```diff
@@ -0,0 +1,128 @@
...
+{
+public:
+  PendulumController(std::chrono::nanoseconds period, PIDProperties pid)
+  : publish_period_(period), pid_(pid),
+    command_message_(std::make_shared<pendulum_msgs::msg::JointCommand>()),
+    message_ready_(false)
+  {
+    command_message_->position = pid_.command;
+    dt_ = publish_period_.count() / (1000.0 * 1000.0 * 1000.0);
+    assert(!isnan(dt_));
```
dirk-thomas: An `assert` is not very string here. If this is an invariant which should always be checked in order for the following code to work correctly it should *always* be checked. I would suggest throwing a `runtime_error` instead in order to also check in non-debug builds.

Same below.

pendulum_control/src/pendulum_controller.hpp
```diff
@@ -0,0 +1,128 @@
...
+    assert(dt_ != 0);
+  }
+
+  // Calculate new command based on new sensor state and PID controller properties
+  void on_sensor_message(const pendulum_msgs::msg::JointState::SharedPtr & msg)
+  {
+    if (!msg) {
+      throw std::runtime_error("received NULL message in callback");
+    }
```
dirk-thomas: Is this even possible? Should rclcpp ever trigger a callback without a message?

pendulum_control/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+
+link_directories(${rttest_LIBRARY_DIR})
+include_directories(${rttest_INCLUDE_DIRS})
```
dirk-thomas: Since you specify `rttest` in `ament_target_dependencies` these two lines are not necessary.

pendulum_control/src/pendulum_controller.hpp
jacquelinekay: well, I've already made this program POSIX-specific by using `pthread` to set thread priority and scheduler type...

pendulum_control/CMakeLists.txt
jacquelinekay: Yes, or `rmw_implementation` (since that package provides `get_available_rmw_implementations`).

pendulum_msgs/msg/JointState.msg
jacquelinekay: I agree that only using static messages for realtime is a huge limitation and I should at least support dynamically sized messages with a bounded size. I actually started a pull request related to this but I put it on the backburner to focus on what the demo would look like.

I think Adolfo is suggesting a way to lock dynamically sized messages during the startup stage of the node lifecycle, so that you can use a preallocated, statically sized buffer for the message during execution. Is that accurate? It seems like this is useful when the size of the messages becomes known when the program starts, but does not change during real-time execution.

pendulum_control/CMakeLists.txt
wjwwood: I'd say `rmw_implementation` too.

dirk-thomas: Since it should use `get_available_rmw_implementations` it must not be in `rmw`. `rmw_implementation` it is.

jacquelinekay: @dirk-thomas are you taking care of this or would you like me to do it?

dirk-thomas: I will write the generic function and then ping you back here.

pendulum_msgs/msg/JointState.msg
adolfo-rt: > I don't think we can make the complete pipeline preallocated here

I don't think it's necessary to preallocate the entire pipeline.

> I think Adolfo is suggesting a way to lock dynamically sized messages during the startup stage of the node lifecycle, so that you can use a preallocated, statically sized buffer for the message during execution. Is that accurate? It seems like this is useful when the size of the messages becomes known when the program starts, but does not change during real-time execution.

Yes, that's what I meant. And that's one of the reasons why node lifecycle is important. I think it's reasonable to assume that message and other data structure sizes can be determined (and pre-allocated) at initialization/configure time, which is non real-time. The active/running state would incur no further allocations. I find it also reasonable to expect a node reconfiguration (something like stop→cleanup→configure→start) to account for a problem size change.

I don't doubt there are very legitimate usecases out there for hard-coded message upper bounds, but I've so far not had the need for them. On the other hand, I can't imagine generic robot control software without (pre-allocated) dynamically sized messages.

pendulum_control/CMakeLists.txt
dirk-thomas: ros2/rmw_implementation#3 adds the function to `rmw_implementation`.

In the following PRs you see how it is being used: ros2/examples#48 ros2/demos#10

jacquelinekay:
To enable CI for this demo, we need to add the dependency rttest to ros2.repos and clean up rttest for Jenkins:

ros2/examples/pull/49
ros2/rttest/pull/2

pendulum_control/src/pendulum_controller.hpp
tfoote: FYI you can get it in Windows too with a define: https://stackoverflow.com/questions/6563810/m-pi-works-with-math-h-but-not-with-cmath-in-visual-studio

pendulum_msgs/msg/JointState.msg
jacquelinekay: I've made an issue for this idea, let's move the discussion there: ros2/ros2/issues/95

jacquelinekay:
I changed the CMakeLists to avoid building this package on OSX and Windows and started CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/251/
http://ci.ros2.org/job/ros2_batch_ci_osx/179/
http://ci.ros2.org/job/ros2_batch_ci_windows/258/

We should expect Linux to build the package cleanly and OSX and Windows to avoid building it (though they will install a dummy "pendulum_control" package, since I don't know how to tell ament to avoid installing the package).

dirk-thomas:
If you don't call `ament_package` it won't install anything. Not sure if that generates any issues down the road though.

jacquelinekay:
ah, I understand my issue now...

If I returned from the package without calling `find_package(ament_cmake)`
beforehand, ament would try to install the package and complain about no
install rule:

```
+++ Installing 'pendulum_control'
==> '. /home/jackie/code/ros2_ws/build/pendulum_control/cmake__install.sh
&& /usr/bin/make install' in
'/home/jackie/code/ros2_ws/build/pendulum_control'
make: *** No rule to make target `install'.  Stop.
```

William suggested I add ament_package as a workaround, but it I just find
ament before returning, it returns without throwing an error or installing
the package.

dirk-thomas:
If not having any install steps is a viable use case we could make `ament_tools` skip the install step in that case.

jacquelinekay:
well, looks like calling `return()` after `find_package(ament_cmake)` but before `ament_package()` still doesn't work on Jenkins:

http://ci.ros2.org/job/ros2_batch_ci_osx/180/
http://ci.ros2.org/job/ros2_batch_ci_windows/260/

Is there a recommended way to conditionally skip the installation of a package?

dirk-thomas:
If a package does not use `install()` anywhere it won't have an *install* target. Currently ament_tools always invokes `make install`. Either we have to update the tool to check if the target exists and only then invoke it or the package must install *something*. I can make a PR for `ament_tools` if we think it is viable to not have an install package.

wjwwood:
I agree that the build tool should not fail if a package has no install target, but calling `ament_package()` should have created one in order to install the package.xml?

jacquelinekay:
@wjwwood The failed build happened after I took out the ament_package workaround, I thought simply calling `find_package(ament_cmake)` before returning would work but it did not.

I think it is viable to allow for conditionally not installing a package, unless we want to discourage people from making code that only compiles on one OS.

The situation here is that `rttest` only compiles on Linux, so for OSX and Windows, an empty package is installed. This means that I have to repeat the conditional `if(APPLE OR WIN_32)` here to avoid attempting to compile on OSX and Windows, instead of skipping if `rttest` was not installed. For cases where conditionally building a package is more complicated, repeating the condition could be very unwieldy.

tfoote:
Overall this looks good to me. +1 pending the CI issues resolution.

pendulum_control/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(pendulum_control)
+find_package(ament_cmake REQUIRED)
```
dirk-thomas: If you move the `find_package()` call below the platform check it will exit a bit fast on non Linux system.

pendulum_control/src/pendulum_controller.hpp
```diff
@@ -0,0 +1,136 @@
...
+  {
+    command_message_->position = pid_.command;
+    dt_ = publish_period_.count() / (1000.0 * 1000.0 * 1000.0);
+    if (isnan(dt_) || dt_ == 0) {
+      throw std::runtime_error("Invalid dt_ calculated in PendulumController constructor");
+    }
+  }
+
+  // Calculate new command based on new sensor state and PID controller properties
+  void on_sensor_message(const pendulum_msgs::msg::JointState::SharedPtr & msg)
```
dirk-thomas: Avoid using references for smart pointers (see ros2/ros2#94).

pendulum_control/src/pendulum_controller.hpp
```diff
@@ -0,0 +1,136 @@
...
+      command_message_->position = 0;
+    }
+
+    if (isnan(command_message_->position)) {
+      throw std::runtime_error("Resulting command was NaN in on_sensor_message callback");
+    }
+    message_ready_ = true;
+  }
+
+  pendulum_msgs::msg::JointCommand::SharedPtr get_next_command_message()
```
dirk-thomas: Can this be const or is the joint command supposed to be modified from the outside?

pendulum_control/src/pendulum_demo.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // Initialization phase
+
+  // TODO(jacquelinekay) runtime tuning of controller properties and physical characteristics
+  pendulum_control::PendulumProperties properties;
+  pendulum_control::PIDProperties pid;
+
+  auto pendulum_motor =
+    std::make_shared<pendulum_control::PendulumMotor>(std::chrono::nanoseconds(970000),
+      properties);
+  auto pendulum_controller = std::make_shared<pendulum_control::PendulumController>(std::chrono::nanoseconds(
```
dirk-thomas: The newline separate `nanoseconds` from the number. it could be wrapped after `PendulumController>(` instead.

pendulum_control/src/pendulum_demo.cpp
```diff
@@ -0,0 +1,189 @@
...
+        960000), pid);
```
dirk-thomas: Wrapped lines should use one indentation level (+2 spaces). Not sure if uncrustify is picking this up though.

Same below.

pendulum_control/src/rtt_executor.hpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  virtual ~RttExecutor() {}
+
+  bool is_rttest_ready() const
+  {
+    return rttest_ready;
+  }
+
+  bool is_running()
```
dirk-thomas: Can this be `const`?

dirk-thomas:
+1 (beside nit picking).

jacquelinekay:
Last CI run (probably):

http://ci.ros2.org/job/ros2_batch_ci_linux/257/
http://ci.ros2.org/job/ros2_batch_ci_osx/185/
http://ci.ros2.org/job/ros2_batch_ci_windows/264/

PR #9, opened 2015-08-17
Use raw data directly

esteve:
Fix compiler warning

Connects to ros2/ros2#53


esteve:
CI job:

http://ci.ros2.org/job/ros2_batch_ci_windows/210/

dirk-thomas:
What happened to the previous patch which did exactly this a week ago??? 38e89cd6b4c8e01284d33ba4eea11c66b8138721

esteve:
@dirk-thomas I don't know, I found it weird because I remembered I already pushed this change, but I just thought it was my faulty memory. Anyway, I just checked the changelog for master and it's not there.

dirk-thomas:
The dashboard history does only show information for the last four days which does not include who force pushed to master.

+1 anyway again to fix the memory leak.

PR #12, opened 2015-08-25
Intra-process image pipeline example

jacquelinekay:
This is a first pass at an example that exercises intra-process communication for an image processing pipeline.

Adds:
* img_pipeline.cpp, a simple main class which loads the nodes into one SingleThreadedExecutor and spins.
* utils.hpp, contains Mat <-> ROS Image.msg conversion functions.
* 4 subclasses of Node which execute intra-process comms by default:
  * DeviceNode, which publishes a video stream to a rostopic.
  * ExtractorNode, which performs a read-only operation on the image it receives (histogram calculation) and republishes the message on a new topic.
  * ModifierNode, which writes to the image (color inversion) and publishes the modified message on a new topic.
  * VisualizerNode, which shows the modified image in an OpenCV window.

Todo:
- [ ] Terminal output that shows that intra-process communication is being exercised (e.g. print out pointer values for messages).
- [ ] Write more main examples exercising different process compositions, or integrate with class_loader.
- [ ] Copy performance issue: currently, `convert_message_to_frame` performs a zero-copy conversion from ROS message to OpenCV Mat "frame" by assigning to Mat data pointer to the start of the ROS Image data vector. However, `convert_frame_to_message` copies the Mat data into the ROS Image data vector, because I couldn't figure out how to directly assign the pointer to the underlying data of a vector (and I don't think this is possible).
- [ ] Framerate is fairly low right now; publishing at higher than ~33Hz causes a race condition similar to ros2/ros2/issues/91 where the publishers for the other nodes do not get initialized before the executor starts spinning.


esteve:
I found this bug in OpenCV on Windows (see #13):

http://stackoverflow.com/questions/20854682/opencv-multiple-unwanted-window-with-garbage-name

replacing `cv::Mat` with `CvMat` (and `imshow` and `waitKey` with `cvShowImage` and `cvWaitKey`, respectively) should do the trick.

intra_process_comms/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ find_package(rclcpp REQUIRED)
 find_package(rmw REQUIRED)
 find_package(rmw_implementation REQUIRED)
 find_package(std_msgs REQUIRED)
+find_package(sensor_msgs REQUIRED)
```
dirk-thomas: Alpha order.

intra_process_comms/include/intra_process_comms/extractor_node.hpp
```diff
@@ -0,0 +1,94 @@
...
+        if (planes_.empty()) {
+          cv::split(*frame_, planes_);
+        }
+        int histSize = 32;
+        float range[] = {0, 256};
+        const float * histRange = {range};
+        // TODO: encoding type
+        cv::calcHist(&planes_[0], 1, 0, cv::Mat(), b_hist_, 1, &histSize, &histRange);
+        cv::calcHist(&planes_[1], 1, 0, cv::Mat(), g_hist_, 1, &histSize, &histRange);
+        cv::calcHist(&planes_[2], 1, 0, cv::Mat(), r_hist_, 1, &histSize, &histRange);
```
dirk-thomas: This should at least check the vector size before starting to access the index.

intra_process_comms/include/intra_process_comms/modifier_node.hpp
```diff
@@ -0,0 +1,76 @@
...
+    bool use_intra_process_comms = true)
+  : Node(node_name, use_intra_process_comms)
+  {
+    rmw_qos_profile_t qos = rmw_qos_profile_default;
+    qos.depth = 16;
+    // after getting the first message, reassign based on encoding type
+    img_pub_ = this->create_publisher<sensor_msgs::msg::Image>("modified_image", qos);
+
+    auto img_sub_callback =
+      [this](sensor_msgs::msg::Image::SharedPtr msg) -> void {
```
dirk-thomas: By receiving the message as a non-const shared pointer it already gets copied here, right?

intra_process_comms/include/intra_process_comms/modifier_node.hpp
```diff
@@ -0,0 +1,76 @@
...
+        convert_message_to_frame(msg, frame_);
+
+        cv::bitwise_not(*frame_, *frame_);
+
+        convert_frame_to_message(*frame_, msg_number_, image_msg_);
```
dirk-thomas: The extra copy into `image_msg_` makes the pipeline not zero-copy (beside the copy commented on above).

I think the approach should be:
* receive a unique pointer (or a non-const shared ptr for now which implies a copy)
* use cv on the loaned data pointer to manipulate it
* publish the ros message again without copying it from cv back to ros (since the data pointer was loaned the original ros message already contains the changed data)

intra_process_comms/CMakeLists.txt
jacquelinekay: k. I'll keep OpenCV in its own category instead of sorting it into alphabetical order, since it's not a ROS package (I assume that's our convention based on https://github.com/ros2/demos/blob/master/image_tools/CMakeLists.txt).

dirk-thomas: Sounds good. With same below I referred to line 33.

intra_process_comms/include/intra_process_comms/modifier_node.hpp
jacquelinekay: I've implemented as much of this approach as I can until William completes support unique pointer publication.

There's still a copy in the first stage of the pipeline from the OpenCV frame received from the camera into the underlying storage vector of the ROS image message. I'm not sure what to do about it.

https://github.com/ros2/demos/pull/12/files#diff-b0f783a56e5774affca5619262fea52eR59

intra_process_comms/include/intra_process_comms/extractor_node.hpp
```diff
@@ -0,0 +1,97 @@
...
+    // after getting the first message, reassign based on encoding type
+    img_pub_ = this->create_publisher<sensor_msgs::msg::Image>("extracted_image", qos);
+
+    //feature_pub_ = this->create_publisher<sensor_msgs::msg::Image>("extracted_features", qos);
+
+    // Subscribe to image
+    auto sub_callback =
+      [this](const sensor_msgs::msg::Image::SharedPtr msg) -> void {
+        image_msg_ = msg;
+        img_pub_->publish(image_msg_);
```
dirk-thomas: Why would the next node in the pipeline subscribe to this relayed topic rather then the original once?

jacquelinekay: @wjwwood suggested that the read-only node republishes the message when we discussed the design of this demo on Monday, though looking at the code now I'm not sure why this is better.

wjwwood: It prevents copying of the message since there are never more than one subscription for each publisher.

wjwwood:
Ok I updated this pr with a new example: https://github.com/ros2/demos/pull/12/files#diff-9c1f8132951cbb58bc4bbd4f97b870a6

Required changes in `rclcpp`, `rmw`, and `system_tests` to follow.

wjwwood:
This is an example of the output from the above linked new example:

```
% ./install/bin/two_node_pipeline
Published message with value: 0, and address: 0x7fc5d5024cb0
Published message with value: 1, and address: 0x7fc5d2e40b10
 Received message with value: 1, and address: 0x7fc5d2e40b10
Published message with value: 2, and address: 0x7fc5d2d3acb0
 Received message with value: 2, and address: 0x7fc5d2d3acb0
Published message with value: 3, and address: 0x7fc5d5011a10
 Received message with value: 3, and address: 0x7fc5d5011a10
Published message with value: 4, and address: 0x7fc5d5011a10
 Received message with value: 4, and address: 0x7fc5d5011a10
Published message with value: 5, and address: 0x7fc5d5011a10
 Received message with value: 5, and address: 0x7fc5d5011a10
Published message with value: 6, and address: 0x7fc5d2d3acb0
 Received message with value: 6, and address: 0x7fc5d2d3acb0
Published message with value: 7, and address: 0x7fc5d5011a10
 Received message with value: 7, and address: 0x7fc5d5011a10
Published message with value: 8, and address: 0x7fc5d5011a10
 Received message with value: 8, and address: 0x7fc5d5011a10
Published message with value: 9, and address: 0x7fc5d5011a10
 Received message with value: 9, and address: 0x7fc5d5011a10
Published message with value: 10, and address: 0x7fc5d5011a10
 Received message with value: 10, and address: 0x7fc5d5011a10
...
```

wjwwood:
Ok, so I've got a pretty nice demo going now. See the files in:

https://github.com/ros2/demos/tree/e9e70bfce41eae53c033d29d49a47df6d1e86224/intra_process_comms/src/image_pipeline

This is just a refactoring/reorganization of what @jacquelinekay did (Thanks for that btw, gave me a huge head start on the CV stuff).

So, now if you run `image_pipeline` you'll get a Window with the image annotated with text. Like this:

screen shot 2015-08-27 at 1 28 16 am

The demo is made up of three nodes at the moment: `camera_node` -> `watermark_node` -> `image_view_node`. Each of the nodes write the address of the pointer they got or are publishing. The watermark node additionally writes some other text. You can press space bar to pause/unpause and you can press `q` or ESC to stop it.

I managed to find a way to create a cv::Mat and operate on it without copying the data out of the message. So the only copying is when I copy the data from the captured image into the original ros message.

I've also added two other toy problems. I still have to write some of the tutorial, but I'll hopefully be able to knock that out quick now that I've got some of the demo done.

intra_process_comms/src/cyclic_pipeline/cyclic_pipeline.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <chrono>
+#include <cstdio>
+
+#include <rclcpp/rclcpp.hpp>
+#include <std_msgs/msg/int32.hpp>
+
+// This node receives an Int32, waits 1 second, then increments and sends it.
+struct IncrementerPipe : public rclcpp::Node
+{
+  IncrementerPipe(std::string name, std::string in, std::string out) : Node(name, true)
```
jacquelinekay: Pass the input strings as `const` references.

wjwwood: cb2cdee

wjwwood:
CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/276/
- http://ci.ros2.org/job/ros2_batch_ci_osx/195/
- http://ci.ros2.org/job/ros2_batch_ci_windows/276/

intra_process_comms/src/image_pipeline/camera_node.hpp
```diff
@@ -0,0 +1,92 @@
...
+    if (!cap_.isOpened()) {
+      throw std::runtime_error("Could not open video stream!");
+    }
+    rmw_qos_profile_t qos = rmw_qos_profile_default;
+    qos.history = RMW_QOS_POLICY_KEEP_LAST_HISTORY;
+    qos.depth = 1;
+
+    pub_ = this->create_publisher<sensor_msgs::msg::Image>(output, qos);
+
+    thread_ = std::thread(std::bind(&CameraNode::loop, this));
```
jacquelinekay: Can you give some justification for using `std::thread` instead of `create_wall_timer` for publishing this message? My guess is that you don't want the image capture to run in the same thread as the executor, but you also don't want to use a MultiThreadedExecutor because we haven't vetted the intra-process stuff for thread safety yet.

Also, I noticed you don't sleep in the `loop` function. Does OpenCV block when we get a new frame? Or is this a busy while loop?

wjwwood: I used a thread because I want the loop to run continuously (without a sleep), so the timer isn't well suited for that. OpenCV will block when you call `frame_ << cap_;`.

wjwwood: Also, I added some thread safety protection on the publishing side:

https://github.com/ros2/rclcpp/pull/98/files#diff-847e0cb63d880a84b709b86a179d2340R106

wjwwood: The other idea I had was to add an "on_idle" callback list for the executor, which is similar to other event loop systems:

- WxWidgets has an idle event which you can handle and do extra work in: http://www.wxpython.org/docs/api/wx.IdleEvent-class.html

- In Qt you can use a Timer with timeout 0 (might work for us too) or processEvent.

http://doc.qt.io/qt-4.8/qtimer.html#details
> As a special case, a QTimer with a timeout of 0 will time out as soon as all the events in the window system's event queue have been processed. This can be used to do heavy work while providing a snappy user interface:

- libuv also has idle processing: https://nikhilm.github.io/uvbook/basics.html#idling

jacquelinekay: Yeah, the mechanism for prioritizing queued work in the Executor is a bit inflexible right now. (related: ros2/rclcpp/issues/76) Another feature to push on after the release...

jacquelinekay:
With regards to getting this demo to work on Windows, I think you can convert to a CvMat without incurring copy using this constructor (it just assigns a data pointer without allocating new memory):

https://github.com/Itseez/opencv/blob/2.4/modules/core/include/opencv2/core/types_c.h#L729

you can do this just before calling `CvShowImage` to fix the bug Esteve found.

esteve:
@jacquelinekay yeah, that's what I did:

https://github.com/ros2/demos/blob/master/image_tools/src/showimage.cpp#L53

wjwwood:
Work around is in place and it compiles on Windows, new CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/283/
- http://ci.ros2.org/job/ros2_batch_ci_osx/208/
- http://ci.ros2.org/job/ros2_batch_ci_windows/283/

wjwwood:
And again:

- http://ci.ros2.org/job/ros2_batch_ci_linux/286/
- http://ci.ros2.org/job/ros2_batch_ci_osx/212/
- http://ci.ros2.org/job/ros2_batch_ci_windows/285/

wjwwood:
You guys are killing me... rebased again.

wjwwood:
One more round of CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/288/
- http://ci.ros2.org/job/ros2_batch_ci_osx/213/
- http://ci.ros2.org/job/ros2_batch_ci_windows/288/

dirk-thomas:
None of the classes is in a proper namespace following the package name.

intra_process_comms/src/two_node_pipeline/two_node_pipeline.cpp
```diff
@@ -0,0 +1,66 @@
...
+  rclcpp::Subscription<std_msgs::msg::Int32>::SharedPtr sub_;
+};
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  rclcpp::executors::SingleThreadedExecutor executor;
+
+  auto producer = std::make_shared<Producer>("number");
+  auto consumer = std::make_shared<Consumer>("number");
```
dirk-thomas: Two nodes should not have the same name.

esteve:
I just rebased this and updated with feedback from @dirk-thomas. The non-rebased version of the branch can be found at https://github.com/ros2/demos/tree/intra_process_img-prerebase if you need it @wjwwood

intra_process_comms/src/two_node_pipeline/two_node_pipeline.cpp
wjwwood: That's not the node name, and so they have different names.

I reverted the change in 58f7339.

dirk-thomas: Oh, it is actually the topic name. That is unexpected to change the constructor argument semantic in a subclass. Good that you catched it.

wjwwood: Yeah, It's not very nice, but at the time I thought it was unnecessary to additionally pass the node name.

wjwwood:
Ok, I've squashed everything and opened up one more set of CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/292/
- http://ci.ros2.org/job/ros2_batch_ci_osx/217/
- http://ci.ros2.org/job/ros2_batch_ci_windows/291/

wjwwood:
Ok, other than the flaky tests and a few compiler warnings on Windows (which I fixed and tested manually) this looks good to go.

dirk-thomas:
Fixing windows warnings in #20

PR #13, opened 2015-08-26
Rewrite OpenCV code in C

esteve:
This is needed to avoid this on Windows:

http://stackoverflow.com/questions/20854682/opencv-multiple-unwanted-window-with-garbage-name

### DIFF
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -125,7 +125,10 @@ int main(int argc, char * argv[])
         convert_frame_to_message(flipped_frame, i, msg);
       }
       if (show_camera) {
-        cv::imshow("cam2image", frame);
+        // NOTE(esteve): Use C version of cvShowImage to avoid this on Windows:
+        // http://stackoverflow.com/questions/20854682/opencv-multiple-unwanted-window-with-garbage-name
+        CvMat cvframe = frame;
+        cvShowImage("cam2image", &cvframe);
         cv::waitKey(1);
       }
       std::cout << "Publishing image #" << i << std::endl;
--- a/image_tools/src/showimage.cpp
+++ b/image_tools/src/showimage.cpp
@@ -48,7 +48,10 @@ void show_image(const sensor_msgs::msg::Image::ConstSharedPtr & msg) {
     msg->height, msg->width, encoding2mat_type(msg->encoding),
     const_cast<unsigned char *>(msg->data.data()), msg->step);
 
-  cv::imshow("view", frame);
+  // NOTE(esteve): Use C version of cvShowImage to avoid this on Windows:
+  // http://stackoverflow.com/questions/20854682/opencv-multiple-unwanted-window-with-garbage-name
+  CvMat cvframe = frame;
+  cvShowImage("showimage", &cvframe);
   cv::waitKey(1);
 }
 
@@ -64,6 +67,8 @@ int main(int argc, char * argv[])
     argc, argv, &depth, &reliability_policy, &history_policy)) {
     return 0;
   }
+
+  cvNamedWindow("showimage", CV_WINDOW_AUTOSIZE);
 
   auto node = rclcpp::node::Node::make_shared("showimage");
 


dirk-thomas:
Does the same apply to #12?

esteve:
@dirk-thomas seems likely, should be easy to migrate #12 before merging it.

tfoote:
+1

image_tools/src/cam2image.cpp
```diff
@@ -125,7 +125,11 @@ int main(int argc, char * argv[])
         convert_frame_to_message(flipped_frame, i, msg);
       }
       if (show_camera) {
-        cv::imshow("cam2image", frame);
+        // NOTE(esteve): Use C version of cvShowImage to avoid this on Windows:
+        // http://stackoverflow.com/questions/20854682/opencv-multiple-unwanted-window-with-garbage-name
+        CvMat cvframe;
```
vrabaud: It's cleaner to use http://docs.opencv.org/modules/core/doc/basic_structures.html#mat-operator-cvmat

image_tools/src/showimage.cpp
```diff
@@ -44,12 +44,16 @@ void show_image(const sensor_msgs::msg::Image::ConstSharedPtr & msg) {
   ss << "Received image #" << msg->header.frame_id << std::endl;
   std::cout << ss.str();
 
-  cv::Mat frame(
-    msg->height, msg->width, encoding2mat_type(msg->encoding),
+  // NOTE(esteve): the OpenCV portions of this program are written in C to avoid
+  // the following bug on Windows:
+  // http://stackoverflow.com/questions/20854682/opencv-multiple-unwanted-window-with-garbage-name
+  CvMat frame;
+  cvInitMatHeader(
```
vrabaud: same, and I would still keep the creation of a cv::Mat for clarity and to make it easier to migrate one day (to OpenCV 3). cv::Mat constructor with a pointer does not take ownership of the data anyway

vrabaud:
I don't understand why this is happening ...
I hope you guys have OpenCV 2.4.11 https://github.com/Itseez/opencv/releases/tag/2.4.11  (I believe there is no Windows installer for that on the download page).
If you have a tiny bit of time to investigate, I'd be curious to know whether it still happens when cvNameWindow is at the beginning.

C++ and C are normally interchangeable as they are replaced by one another (depending on the case).

jacquelinekay:
I'm a bit confused about the solution because the SO link you sent has no mention of the C API, but their solution was to add a magical preprocessor definition to the file. How did you arrive at this solution?

esteve:
@jacquelinekay at first I used the suggestion posted on SO on a simple test program, but it breaks linkage on our binaries, as all symbols must be compiled with `_ITERATOR_DEBUG_LEVEL=0`. Then I figured out there must be something wrong in how `std::string` is encoded or interpreted in Windows/OpenCV, so given that the preprocessor flag just affects checked iterators, I opted to use the lowlevel C version of `imshow` and `namedWindow` which accept a `const char *` instead.

esteve:
@vrabaud thanks for your feedback. Yeah, this happens with OpenCV 2.4.11. I had to build it myself because we're using VS2015 (vc14), which is not available in the official binaries. I don't know if the official binaries are affected or it's just my setup, I can dig a bit deeper, but it'll have to wait until next week.

jacquelinekay:
it could be an issue with `c_str()` conversion, since that's the only difference between `cv::namedWindow`:

https://github.com/Itseez/opencv/blob/2.4.11/modules/highgui/src/window.cpp#L156

https://github.com/Itseez/opencv/blob/2.4.11/modules/highgui/src/window.cpp#L257

vrabaud:
if it's in c_str(), can you actually narrow down by using ```cv::NamedWindow``` and not ```cvNamedWindow```

vrabaud:
FYI: http://www.codeexplain.com/code/a3fce855-78f7-4467-8aa5-81fdc2dc2fbb

esteve:
I also tested this encoding (http://en.cppreference.com/w/cpp/language/string_literal) the title string in UTF8, UTF16, etc. (thanks @jacquelinekay), and it still doesn't work. The only way this works fine is with `cvShowImage` and `cvNamedWindow`

PR #14, opened 2015-08-26
Document pendulum control demo and add malloc hook

jacquelinekay:
Add lots of comments and correct the README.

With the help of @codebot, I also added a `malloc_hook` to show where allocations happen in this program. I am happy to report that no mallocs occur during `executor::spin`.

@gerkey to test if he has time.

### DIFF
--- a/pendulum_control/README.md
+++ b/pendulum_control/README.md
@@ -1,23 +1,31 @@
 ## Build instructions
 
-If you haven't already, clone [rttest](https://github.com/jacquelinekay/rttest) into your ament workspace.
+If you haven't already, clone [rttest](https://github.com/ros2/rttest) into your ament workspace.
 
-Build:
+Build your ament workspace:
 
 ```
-./src/ament/ament_tools/scripts/ament.py build --build-tests --symlink-install --only pendulum_control
+./src/ament/ament_tools/scripts/ament.py build --symlink-install
 ```
+You can add `--only pendulum_control` if you want to only rebuild this package.
 
 Run:
 
 ```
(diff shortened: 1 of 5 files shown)


tfoote:
[Line 3](https://github.com/ros2/demos/pull/14/files#diff-7cc1c7220d23fb825f6ae8d3bfe24af3R3) of the Readme should reference the new repo url. It redirects for now until you make a local fork.

pendulum_control/README.md
```diff
@@ -19,6 +19,12 @@ The demo should work with both middleware implementations.
 A few command line arguments related to real-time performance profiling are provided by rttest.
 See https://github.com/ros2/rttest/blob/master/README.md for more information.
 
+The demo will spit out a lot of output as it runs. Try redirecting the results to a file:
+
+```
+./build/pendulum_control/pendulum_demo[__rmw_opensplice/__rmw_connext] > output.txt
```
tfoote: After sourcing install/setup.* you should use the executables in install/bin which is on the path already. The ones in build are intermediate products only. 

Also there's no need to build the unit tests for the demo. It will save time compiling. And I wouldn't recommend using `--only pendulum_control` since you do not know the state of the workspace and whether every dependency has already been compiled. (I made a new workspace and didn't have anything built yet).

wjwwood: You can use `--end-with` instead.

dirk-thomas:
Same comment for docblocks as in https://github.com/ros2/rclcpp/pull/97#issuecomment-135483396

PR #15, opened 2015-08-27
Comment QoS demo

jacquelinekay:
I had a few questions while reading this code, perhaps @esteve can answer them:

* Who publishes to the "flip_image" topic?
* Should we add `ament_lint_auto` to this package to enable linting in Jenkins?
* Do we need to sleep in the while loop in `cam2image`? In the latest version of ros2/demos/pull/12, the thread that gets the image from OpenCV does not sleep, because getting the frame from the video capture is a blocking operation. Should we do things the same way here or would you prefer to throttle the update rate?

### DIFF
--- a/image_tools/include/image_tools/options.hpp
+++ b/image_tools/include/image_tools/options.hpp
@@ -20,12 +20,34 @@
 
 #include <rmw/types.h>
 
+/// Find "option" in the argument vector.
+/**
+ * \param[in] args The argument vector
+ * \param[in] option The option to search for
+ * \return True if option was found in args, false otherwise.
+ */
 bool find_command_option(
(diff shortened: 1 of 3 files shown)


dirk-thomas:
The `flip_image` topic is only used by the bridge demo.

esteve:
Good idea about adding `ament_lint_auto`, I'll have a look into it.

I didn't write the code for `cam2image`, but my guess is that the sleep is there to lock the frames per second to 30, but I could be wrong.

esteve:
Just submitted this to add support for `ament_lint_auto`

https://github.com/ros2/demos/pull/16

esteve:
Just rebased this on top of master. Changes look good to me, merging now.

@jacquelinekay thanks!

PR #18, opened 2015-08-28
make producer / consumer plugins in a shared library and load them with pluginlib

dirk-thomas:
Simple example using `pluginlib` (https://github.com/ros/pluginlib/tree/ros2) and indirectly `class loader` (https://github.com/ros/class_loader/tree/ros2).

Before this can be considered it must be possible to build a shared library which only depends on `rclcpp` but not a specific rmw implementation. Currently `rclcpp` depends directly on the headers of a specific implementation:
- https://github.com/ros2/rclcpp/blob/65300405bef94cc391253083f7140041c69c737f/rclcpp/include/rclcpp/node.hpp#L27
- https://github.com/ros2/rclcpp/blob/65300405bef94cc391253083f7140041c69c737f/rclcpp/include/rclcpp/node_impl.hpp#L30-L31

### DIFF
--- a/intra_process_comms/CMakeLists.txt
+++ b/intra_process_comms/CMakeLists.txt
@@ -7,6 +7,7 @@ if(NOT WIN32)
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(pluginlib REQUIRED)
 find_package(rclcpp REQUIRED)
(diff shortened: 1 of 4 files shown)


dirk-thomas:
This has been implemented in ros2/demos#84.

PR #21, opened 2015-08-31
Intra process demo fixups

wjwwood:
I have some fix ups and some additions to the demos:

- New layout of the image pipeline which forks.
- Added PID to the image's so you can see when different processes are involved.
- Use shared_ptr on subscription in a few places.
- Added stand along nodes for each of the nodes so people can create inter process forks in the pipeline themselves.
- Fixed the name of the package to match the folder name.
- Added inline comments.


wjwwood:
CI:

- http://ci.ros2.org/job/ros2_batch_ci_linux/298/
- http://ci.ros2.org/job/ros2_batch_ci_osx/228/
- http://ci.ros2.org/job/ros2_batch_ci_windows/298/

wjwwood:
New Windows job: http://ci.ros2.org/job/ros2_batch_ci_windows/299/

wjwwood:
CI lgtm, and changes are isolated to the demo. I'm going to merge it, please comment and I can fix up or roll back if needed.

tfoote:
I tested the cyclic successfully. 

The all in one ran but gave a bunch of warnings on opensplice, and segfaulted on both connext implementations:

```
tfoote@yeti:~/work/ros2/test_source/install Last: [0] (0s Seconds)
$ image_pipeline_all_in_one__rmw_connext_dynamic_cpp 
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
COMMENDSrWriterService_write:!write. Reliable large data requires asynchronous writer.
PRESPsWriter_writeInternal:!srw->write
  what():  failed to publish message: failed to write, at /home/tfoote/work/ros2/test_source/src/ros2/rmw_connext/rmw_connext_dynamic_cpp/src/functions.cpp:1534
Aborted (core dumped)
tfoote@yeti:~/work/ros2/test_source/install Last: [134] (1s Seconds)
$ image_pipeline_all_in_one__rmw_connext_cpp 
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
COMMENDSrWriterService_write:!write. Reliable large data requires asynchronous writer.
PRESPsWriter_writeInternal:!srw->write
  what():  failed to publish message: failed to publish message, at /home/tfoote/work/ros2/test_source/src/ros2/rmw_connext/rmw_connext_cpp/src/functions.cpp:852
Aborted (core dumped)
tfoote@yeti:~/work/ros2/test_source/install Last: [134] (1s Seconds)
$ image_pipeline_all_in_one__rmw_opensplice_cpp 
========================================================================================
Report      : WARNING
Date        : Mon Aug 31 12:30:41 PDT 2015
Description : using network interface eth0 (172.23.3.99) selected arbitrarily from: eth0, wlan0, docker0

Node        : yeti
Process     : image_pipeline_all_in_one__rmw_opensplice_cpp <23008>
Thread      : ddsi2 7f97a443e700
Internals   : V6.4.140407OSS///ddsi2/q_nwif.c/668/0/1441049441.241554332
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
VIDIOC_QUERYMENU: Invalid argument
^Csignal_handler(2)
tfoote@yeti:~/work/ros2/test_source/install Last: [0] (18s Seconds)
$ 

```

wjwwood:
I don't get any output when running with OpenSplice. I'll have to rebuild with Connext on to test that. I'll report back when I can test it there.

wjwwood:
This seems to be the cause of the Connext error:

https://community.rti.com/kb/what-error-reliable-large-data-requires-asynchronous-write

Though mine doesn't segfault, rather it exits with an error:

```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: failed to publish message: failed to publish message, at /Users/william/ros2/src/ros2/rmw_connext/rmw_connext_cpp/src/functions.cpp:852
```

wjwwood:
I'm not sure what we should do here. I guess I can change the demo to be best effort, but that just avoids the issue. I'm not sure what `ASYNCHRONOUS_PUBLISH_MODE_QOS` will change, so it's not clear that this is the general solution. I suppose we could check for this condition and then change the mode when we have to, but then you have piecewise behavior based on message size, which is surprising to the user.

wjwwood:
Switching to best effort avoids the error. Should I just go for that and open an issue to track the problem?

@esteve did you run into this problem with your demo?

gerkey:
+1 to switching to best effort for now and ticketing it for further investigation.

tfoote:
According to [the docs here](https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/ASYNCHRONOUS_PUBL_Qos.htm) it seems to basically pass the actual publishing to another thread.

wjwwood:
Yeah, that's what I assumed based on the name, but I don't know what affect that will have system wide.

I'm reading about it now, but I think the easiest thing to do is to make it best effort for now and address it later. I can't imagine, however, how this will not affect @esteve's demo when using reliable to transport images. So it might be side-step for my demo, but his will probably continue to have a problem.

wjwwood:
Looks like we could turn on async and then call `wait_for_asynchronous_publishing` (http://community.rti.com/rti-doc/510/ndds/doc/html/api_cpp/classDDSDataWriter.html#ad3c4be7eac0f7fa4aebb04799bfcba45) before returning from publish to keep the same behavior.

esteve:
For my tests I used low resolution images, using HD definitely choked the publisher.

wjwwood:
It's not "choking", it produces an error and crashes. Is that what you mean? I'm only using 320x240, not exactly HD...

PR #25, opened 2015-09-28
use sensor profile since the message might come from the bridge which needs handle large and therefore can't use the default profile

dirk-thomas:
Connect to ros2/ros1_bridge#12


dirk-thomas:
Please review.

tfoote:
+1

PR #26, opened 2015-09-30
Improvements for realtime demo

jacquelinekay:
Add non-realtime safe process that publish and subscribe to real-time nodes and a launch script.

Connects to ros2/rttest#8


pendulum_control/include/pendulum_control/pendulum_controller.hpp
```diff
@@ -96,6 +96,12 @@ class PendulumController
     message_ready_ = true;
   }
 
+  void on_pendulum_setpoint(const pendulum_msgs::msg::JointCommand::SharedPtr msg)
+  {
+    set_command(msg->position);
+    std::cout << "Pendulum set to: " << msg->position << std::endl;
```
dirk-thomas: `printf`

pendulum_control/include/pendulum_control/rtt_executor.hpp
```diff
@@ -70,6 +70,19 @@ class RttExecutor : public executor::Executor
...
   }
 
+  void set_rtt_results_message(pendulum_msgs::msg::RttestResults::SharedPtr msg) const {
+    msg->cur_latency = last_sample;
+    msg->mean_latency = results.mean_latency;
+    msg->min_latency = results.min_latency;
+    msg->max_latency = results.max_latency;
+    msg->minor_pagefaults = results.minor_pagefaults;
+    msg->major_pagefaults = results.major_pagefaults;
+    struct timespec curtime;
```
dirk-thomas: Is the `struct` keyword needed here?

pendulum_control/src/pendulum_demo.cpp
```diff
@@ -264,6 +206,22 @@ int main(int argc, char * argv[])
...
+  // Create a lambda function to accept user input to command the pendulum
+  auto controller_command_callback =
+    [&pendulum_controller](const pendulum_msgs::msg::JointCommand::SharedPtr msg) -> void
+    {
+      pendulum_controller->on_pendulum_setpoint(msg);
+    };
+
+  auto setpoint_sub = controller_node->create_subscription<pendulum_msgs::msg::JointCommand>
+      ("pendulum_setpoint", qos_profile, controller_command_callback, nullptr, false,
+      setpoint_msg_strategy);
```
dirk-thomas: The indentation looks wrong.

pendulum_control/src/pendulum_demo.cpp
```diff
@@ -293,12 +251,25 @@ int main(int argc, char * argv[])
...
+
   // Add a timer to enable regular publication of sensor messages.
   auto motor_publisher_timer = motor_node->create_wall_timer
       (pendulum_motor->get_publish_period(), motor_publish_callback);
   // Add a timer to enable regular publication of command messages.
   auto controller_publisher_timer = controller_node->create_wall_timer
       (pendulum_controller->get_publish_period(), controller_publish_callback);
+  // Add a timer to enable regular publication of results messages.
+  auto logger_publisher_timer = controller_node->create_wall_timer
+      (logger_publisher_period, logger_publish_callback);
```
dirk-thomas: Usually we keep the open parenthesis at the end of the previous line to make the first line more readable. (The two statements above use the same wrapping.)

pendulum_msgs/msg/RttestResults.msg
```diff
@@ -0,0 +1,11 @@
+builtin_interfaces/Time stamp
+
+pendulum_msgs/JointCommand command
+pendulum_msgs/JointState state
```
dirk-thomas: You don't have to mention the same package name explicitly. I am not sure which option we should prefer:
* either skip it when possible for simplifying the file and avoiding repetition of the package name
* always use fully qualified names including the package name to be explicit

pendulum_msgs/msg/RttestResults.msg
```diff
@@ -0,0 +1,11 @@
+builtin_interfaces/Time stamp
```
dirk-thomas: What semantic does the `stamp` have? Start time? End time?

pendulum_msgs/msg/RttestResults.msg
jacquelinekay: Yeah, I did this just because it seems to be used other places in the message definitions when a sub-message from out of the package is being used (like builtin_interfaces). I think I'll omit the package name from messages that are included in this package but not messages that come from outside of the package.

pendulum_msgs/msg/RttestResults.msg
jacquelinekay: It is the time at which the result represented by this message was taken.

pendulum_control/CMakeLists.txt
```diff
@@ -30,6 +32,28 @@ add_executable_for_each_rmw_implementations(pendulum_demo
...
+  TARGET_DEPENDENCIES
+  "rclcpp"
+  "pendulum_msgs"
+  INSTALL
+)
+
+#install(
+#  DIRECTORY scripts/
+#  DESTINATION bin
+#)
```
dirk-thomas: I would suggest to either install it or remove the comment all together.

pendulum_control/scripts/pendulum_launch.bash
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+pendulum_logger__rmw_opensplice_cpp &
+logger_pid=$!
+
+# run pendulum demo indefinitely
+pendulum_demo__rmw_opensplice_cpp -i 0 &
+#pendulum_demo__rmw_opensplice_cpp -i 0
```
dirk-thomas: Are you using these commented command frequently for testing? If not I would suggest to remove them (same for last line in this file).

pendulum_control/src/pendulum_demo.cpp
```diff
@@ -264,6 +200,22 @@ int main(int argc, char * argv[])
       ("pendulum_sensor", qos_profile, controller_subscribe_callback,
       nullptr, false, state_msg_strategy);
 
+  // Create a lambda function to accept user input to command the pendulum
+  auto controller_command_callback =
+    [&pendulum_controller](const pendulum_msgs::msg::JointCommand::SharedPtr msg) -> void
```
dirk-thomas: Do you want a const SharedPtr here or a SharedPtr to a const message?

pendulum_control/src/pendulum_logger.cpp
```diff
@@ -0,0 +1,83 @@
...
+      std::ofstream fstream;
+      fstream.open(filename, std::ios_base::app);
+      struct timespec timestamp;
+      timestamp.tv_sec = msg->stamp.sec;
+      timestamp.tv_nsec = msg->stamp.nanosec;
+      fstream << i << " " << timespec_to_long(&timestamp) <<
+        " " << msg->cur_latency << " " <<
+        msg->minor_pagefaults << " " <<
+        msg->major_pagefaults << std::endl;
+      fstream.close();
```
dirk-thomas: Is it efficient to open / close this file on every callback? Wouldn't it be better to keep the filehandle open?

dirk-thomas:
Beside the small comment this LGTM.

pendulum_control/src/pendulum_demo.cpp
jacquelinekay: Hmm, I would prefer a `ConstSharedPtr` (SharedPtr to const message) but I ran into this issue:

https://github.com/ros2/rclcpp/issues/118

Ignoring the part about publishing const messages for now, this pull request is needed to get the message as a shared pointer to a const in the subscripton callback:

https://github.com/ros2/rclcpp/pull/119

PR #33, opened 2015-10-27
remove HeapPoolMemoryStrategy

jacquelinekay:
I will change the memory strategy back when there's an acceptable replacement that uses the new MemoryStrategy API.

Connects to ros2/rclcpp#137


esteve:
+1

PR #35, opened 2015-10-28
Fix cpplint warnings

esteve:
Fix cpplint warnings

Connects to ros2/ros2#12


esteve:
@jacquelinekay gave her thumbs up offline

PR #38, opened 2015-10-30
Fix build for making rclcpp a library

jacquelinekay:
Connects to ros2/rclcpp#140

### DIFF
--- a/pendulum_control/include/pendulum_control/rtt_executor.hpp
+++ b/pendulum_control/include/pendulum_control/rtt_executor.hpp
@@ -17,6 +17,7 @@
 
 #include <cassert>
 #include <cstdlib>
+#include <cstring>
 #include <memory>
 #include <vector>
 


pendulum_control/include/pendulum_control/rtt_executor.hpp
```diff
@@ -15,6 +15,8 @@
 #ifndef PENDULUM_CONTROL__RTT_EXECUTOR_HPP_
 #define PENDULUM_CONTROL__RTT_EXECUTOR_HPP_
 
+#include <string.h>
```
esteve: Since this is a C++ header, I think `#include <cstring>` would be more appropriate.

jacquelinekay: ok, I've changed this. I pushed the same fix to rclcpp last night, so I'll change it in that pull request too. Warning @wjwwood that I'm pushing to his branch.

esteve:
+1

Issue #19, opened 2015-08-29
Windows binaries are built with debug symbols


esteve:

esteve:
![dll_error](https://cloud.githubusercontent.com/assets/33620/9559890/43f75736-4db2-11e5-9602-5df976763c92.png)

esteve:
Installing the VC++ redistributable didn't fix this:

https://www.microsoft.com/en-us/download/details.aspx?id=48145

mikaelarguedas:
Closing this because we haven't faced this in any of the previous release testing of windows binaries

Issue #17, opened 2015-08-28
Tutorial for QoS demo


esteve:

esteve:
@jacquelinekay started this document:

https://github.com/ros2/ros2/wiki/Quality-Of-Service

PR #16, opened 2015-08-28
Add support for ament_lint_auto

esteve:

Issue #11, opened 2015-08-24
pendulum_control: fails to build


vmayoral:
@jacquelinekay this package doesn't build:
```
-- Build files have been written to: /home/victor/ros2_ws/build/pendulum_control
==> '. /home/victor/ros2_ws/build/pendulum_control/cmake__build.sh && /usr/bin/make -j3 -l3' in '/home/victor/ros2_ws/build/pendulum_control'
[ 50%] [100%] Building CXX object CMakeFiles/pendulum_demo__rmw_opensplice_cpp.dir/src/pendulum_demo.cpp.o
Building CXX object CMakeFiles/pendulum_demo.dir/src/pendulum_demo.cpp.o
Linking CXX executable pendulum_demo
Linking CXX executable pendulum_demo__rmw_opensplice_cpp
/usr/bin/ld: CMakeFiles/pendulum_demo__rmw_opensplice_cpp.dir/src/pendulum_demo.cpp.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [pendulum_demo__rmw_opensplice_cpp] Error 1
make[1]: *** [CMakeFiles/pendulum_demo__rmw_opensplice_cpp.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
/usr/bin/ld: CMakeFiles/pendulum_demo.dir/src/pendulum_demo.cpp.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [pendulum_demo] Error 1
make[1]: *** [CMakeFiles/pendulum_demo.dir/all] Error 2
make: *** [all] Error 2

```
Tried adding `-lpthread ` https://github.com/erlerobot/demos/commit/ea77763e595513b256b2075f9afa5428e906339a but doesn't seem to help. Any ideas?

jacquelinekay:
Hmm.

Did you clone https://github.com/ros2/rttest into your workspace? That might provide the missing symbols for `pthread`.

vmayoral:
rttest is there yeap. Bumped with it while following https://github.com/ros2/examples/wiki (did it in a fresh trusty chroot).

[...]

PR #39, opened 2015-11-03
Fix pendulum segfault in binaries

jacquelinekay:
I am still unable to reproduce the segfault observed by @dirk-thomas in the source build, and @tfoote was unable to reproduce it in docker. However, after reproducing it in the binaries and observing that the segfault originates in the exception thrown in the `malloc_hook`, I realized two problems with the implementation:

1. The `malloc_hook` function needs to restore the system malloc to `__malloc_hook` before trying to malloc within the function, else it will infinitely call the custom `malloc_hook`. This happens when the exception is thrown because the exception will try to allocate memory.
2. Throwing an exception within a C function, which is generally bad, and especially bad because malloc is usually called from global `new`, a C++ function that is marked noexcept.

I have no idea why the exception is never seen in a source build, even a source build with RelWithDebInfo build type (which is what the binaries use).

@dirk-thomas, since your Linux VM appears to most consistently reproduce the problem, can you download the artifact from the packaging job for this branch and test this change?

http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/38/

### DIFF
--- a/pendulum_control/src/pendulum_demo.cpp
+++ b/pendulum_control/src/pendulum_demo.cpp
@@ -53,19 +53,14 @@ static void * (* prev_malloc_hook)(size_t, const void *);
 static void * testing_malloc(size_t size, const void * caller)
 {
   (void)caller;
-  /*
-  // Maximum depth we are willing to traverse into the stack trace.
-  static const int MAX_DEPTH = 2;
-  // Instantiate a buffer to store the traced symbols.
-  void * backtrace_buffer[MAX_DEPTH];
-  */
-  if (running) {
-    throw std::runtime_error("Called malloc during realtime execution phase!");
-  }
-
   // Set the malloc implementation to the default malloc hook so that we can call it implicitly
   // to initialize a string, otherwise this function will loop infinitely.
   __malloc_hook = prev_malloc_hook;
+
+  if (running) {
+    fprintf(stderr, "Called malloc during realtime execution phase!\n");
+    exit(-1);
+  }
 
   // Execute the requested malloc.
   void * mem = malloc(size);


dirk-thomas:
Again sometimes it works, sometimes it throws:

    Couldn't set scheduling priority and policy: Operation not permitted
    mlockall failed: Cannot allocate memory
    Couldn't lock all cached virtual memory.
    Pagefaults from reading pages not yet mapped into RAM will be recorded.
    Initial major pagefaults: 0
    Initial minor pagefaults: 721
      what():  Called malloc during realtime execution phase!
    Aborted (core dumped)

The stacktrace within `gdb`:

    #0  0x00007ffff60f8cc9 in __GI_raise (sig=sig@entry=6)
        at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
    #1  0x00007ffff60fc0d8 in __GI_abort () at abort.c:89
    #2  0x00007ffff6a03535 in __gnu_cxx::__verbose_terminate_handler() ()
       from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
    #3  0x00007ffff6a016d6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
    #4  0x00007ffff6a01703 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
    #5  0x00007ffff6a01922 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
    #6  0x0000000000428d57 in testing_malloc (size=<optimized out>, caller=<optimized out>)
        at /home/rosbuild/ci_scripts/workspace/src/ros2/demos/pendulum_control/src/pendulum_demo.cpp:68
    #7  0x00007ffff5be0d72 in c_tableRemove () from /usr/lib/libddskernel.so
    #8  0x00007ffff5be1240 in c_find () from /usr/lib/libddskernel.so
    #9  0x00007ffff5c1c5a9 in v_partitionAdminExists () from /usr/lib/libddskernel.so
    #10 0x00007ffff5bf395d in v_subscriberConnectNewGroup () from /usr/lib/libddskernel.so
    #11 0x00007ffff5bf3a76 in ?? () from /usr/lib/libddskernel.so
    #12 0x00007ffff5c337ea in c_setWalk () from /usr/lib/libddskernel.so
    #13 0x00007ffff5c1d0f6 in v_participantConnectNewGroup () from /usr/lib/libddskernel.so
    #14 0x00007ffff5bd67bb in ?? () from /usr/lib/libddskernel.so
    #15 0x00007ffff5b66aee in ?? () from /usr/lib/libddskernel.so
    #16 0x00007ffff5c303d9 in c_iterWalk () from /usr/lib/libddskernel.so
    #17 0x00007ffff5bb9d50 in ?? () from /usr/lib/libddskernel.so
    #18 0x00007ffff5bcba26 in ?? () from /usr/lib/libddskernel.so
    #19 0x00007ffff6caf182 in start_thread (arg=0x7fffcf891700) at pthread_create.c:312
    #20 0x00007ffff61bc47d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

pendulum_control/src/pendulum_demo.cpp
```diff
@@ -59,14 +59,16 @@ static void * testing_malloc(size_t size, const void * caller)
...
-  if (running) {
-    throw std::runtime_error("Called malloc during realtime execution phase!");
-  }
 
   // Set the malloc implementation to the default malloc hook so that we can call it implicitly
   // to initialize a string, otherwise this function will loop infinitely.
   __malloc_hook = prev_malloc_hook;
 
+  if (running) {
+    fprintf(stderr, "Called malloc during realtime execution phase!");
```
dirk-thomas: Missing `\n`.

gerkey:
Given that this change converted the `throw()` into `fprintf()`, I'm confused as to how you're seeing an exception being thrown.

Btw, we figured out yesterday that, because the new allocator support is incomplete, this demo is now expected to fail by allocating memory at runtime.  It won't work properly again until we implement a real-time-safe allocator, to be done in alpha3.  In the meantime, we're just trying to get to a situation in which the check for real-time-safe behavior behaves correctly and repeatably.

jacquelinekay:
I ran the packaging job right before the changing the throw to an `exit()`.

gerkey:
Oh, that makes sense.  Then the fact that @dirk-thomas is now seeing an exception is a good thing: it demonstrates that moving the `throw()` to happen after swapping back in the real `malloc()` had the intended effect, which is remove the infinite recursion in our  `malloc_hook()`.

The fact that it's not throwing on every run is a bit surprising, but perhaps can be chalked up to the vagaries of how `new` happens to reuse previously allocated memory on any given run?

jacquelinekay:
I could believe that there are some codepaths through `new` that call malloc and some codepaths that cache previously new'ed memory. The fact that Dirk sees this more frequently on the VM makes sense because it probably has a much smaller virtual memory space than our desktops or laptops. Still don't understand why we don't see it in source builds.

The exit should prevent the core dump from happening, although to prevent memory leaks on exit I should rework this to call a function to free memory before `exit`.

jacquelinekay:
starting a new packaging job: http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/39/

jacquelinekay:
again, with the `locale_packaging` fix: http://ci.ros2.org/view/packaging/job/ros2_packaging_linux/40/

dirk-thomas:
> The fact that Dirk sees this more frequently on the VM makes sense because it probably has a much smaller virtual memory space than our desktops or laptops.

My VM has 4 GB of dedicated memory which should be more than enough.

PR #40, opened 2015-11-04
Initialize durability so that the demo does not crash on Windows

esteve:
Somewhere in our code, we don't check that the values in the QoS structure are valid, which causes Windows to crash.

### DIFF
--- a/image_tools/src/showimage.cpp
+++ b/image_tools/src/showimage.cpp
@@ -89,7 +89,7 @@ int main(int argc, char * argv[])
   auto node = rclcpp::node::Node::make_shared("showimage");
 
   // Set quality of service profile based on command line options.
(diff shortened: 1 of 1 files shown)


tfoote:
+1

esteve:
Filed this ticket so that we address validating the QoS profile structure after alpha 2:

https://github.com/ros2/ros2/issues/142

dirk-thomas:
Making the `showimage` node use the default profile will prohibit to use it with the bridge. Therefore it should use the sensor profile instead.

esteve:
This just uses the default profile for the durability setting. But the rest of the settings are either taken from the command line, or use the defaults from https://github.com/ros2/demos/blob/master/image_tools/src/showimage.cpp#L76

PR #42, opened 2015-11-05
Fix pendulum demo regressions for rclcpp_library

jacquelinekay:
I have no idea why leaving out the namespace in these files worked before.

@wjwwood 

Connects to ros2/ros2#147


jacquelinekay:
OMG this is the 42nd pull request in this repo!

wjwwood:
lgtm, should we do CI?

Issue #43, opened 2015-11-17
cover all demos with unit tests


tfoote:
Follow up to ros2/ros2#143

codebot:
burger mode covered some of the image demo(s) :hamburger:

mikaelarguedas:
everlasting issue, closing in favor of more precise tickets

PR #48, opened 2015-12-08
Added test for cam2image and showimage

esteve:
Connects to #43

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+image_tools/test/test_capture.mp4 filter=lfs diff=lfs merge=lfs -text
--- a/image_tools/CMakeLists.txt
+++ b/image_tools/CMakeLists.txt
@@ -50,6 +50,47 @@ call_for_each_rmw_implementation(targets GENERATE_DEFAULT)
 if(AMENT_ENABLE_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  if(NOT WIN32)
+    set(test_executable_extension "")
+    set(test_executable_subfolder "")
+  else()
+    set(test_executable_extension ".exe")
+    set(test_executable_subfolder "/${CMAKE_BUILD_TYPE}")
+    if("${test_executable_subfolder}" STREQUAL "/")
+      set(test_executable_subfolder "/Debug")
+    endif()
+  endif()
+
+  # get the rmw implementations ahead of time
+  get_available_rmw_implementations(middleware_implementations)
+  foreach(middleware_impl ${middleware_implementations})
+    find_package("${middleware_impl}" REQUIRED)
+  endforeach()
+
+  set(RCLCPP_DEMO_SHOWIMAGE_EXPECTED_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/test/showimage")
+  set(RCLCPP_DEMO_CAM2IMAGE_EXPECTED_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/test/cam2image")
+  set(RCLCPP_DEMO_CAM2IMAGE_VIDEO "${CMAKE_CURRENT_SOURCE_DIR}/test/test_capture.mp4")
+
+  foreach(middleware_impl ${middleware_implementations})
+    # get typesupport of rmw implementation to include / link against the corresponding interfaces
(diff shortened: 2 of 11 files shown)


esteve:
CI jobs:

http://ci.ros2.org/job/ci_linux/684/
http://ci.ros2.org/job/ci_osx/560/
http://ci.ros2.org/job/ci_windows_connext_dynamic/34/
http://ci.ros2.org/job/ci_windows_connext_static/34/
http://ci.ros2.org/job/ci_windows_opensplice/752/

esteve:
The added mp4 file is this one, the license should be fine.

https://archive.org/details/SaucersInTheSky_521

dirk-thomas:
Please don't add a 9 MB test file to the repo but download it from somewhere.

esteve:
@dirk-thomas I considered that, but ended up opting for having the test fixtures bundled with the tests, because I like tests that are more or less self contained and because I think there's is value in distributing the whole test. I can see advantages for both approaches (include the test resources or fetching them from a remote location). Thoughts? @codebot @gerkey @jacquelinekay @tfoote @wjwwood

wjwwood:
:-1: :cookie: for including binary files in the repository, as they persist in the history for ever making the repository huge. I'd prefer downloading the video file. Does the content of the video matter? You could instead generate a movie file using OpenCV fairly easily.

wjwwood:
:meat_on_bone: It depends on how you're testing it, I guess it helps because the size of the image changes, but it's not clear to me how we're asserting that the image is displayed in the GUI correctly. It would be useful for acceptance testing but not necessarily automated testing.

tfoote:
Try [github LFS](https://github.com/blog/1986-announcing-git-large-file-storage-lfs) otherwise we can fallback to new cmake download tools from download.ros.org

esteve:
I've tried Git LFS and seems to work, I've triggered a Linux CI job, let's see if this works on the buildfarm:

http://ci.ros2.org/job/ci_linux/688/

:ice_cream:

esteve:
The LFS client needs to be installed on the slaves, I'm going to try adding it to the Docker bootstrap script for Linux and install the client manually on Windows and OSX.

dirk-thomas:
Since we also need to support other VCS should we just go for a download solution and not do any extra stuff for LFS?

esteve:
I'd rather use LFS since it's very well integrated into GitHub.

wjwwood:
> Since we also need to support other VCS should we just go for a download solution and not do any extra stuff for LFS?

Why do we need to support other VCS? And how does that impact the decision to use LFS here?

dirk-thomas:
If we want ROS users to be able to store their code in non-git repos and also have tests requiring large data we need to provide a mechanism to download these data. If we have the infrastructure anyway I don't see a reason to spend extra effort on this special solution.

gerkey:
As we discussed in today's meeting, we're going to try LFS here and see how it works.  That does not stop us from trying other options later.

wjwwood:
I thought the idea is that there is no extra effort? Looks to me that it is a "symlink" file and that you just need to run a command after checkout like initializing submodules. It shouldn't require any extra infrastructure on our part, unless I'm missing something.

Also I thought that CMake provides download infrastructure and we don't need to? So we're not do this instead of spending time on new infrastructure that others would use.

esteve:
@wjwwood you just need to install `git-lfs` and run `git lfs install` only once to. After that, it's all transparent and any checkout will retrieve the files that are tracked via LFS.

esteve:
The Linux CI job passes:

http://ci.ros2.org/job/ci_linux/690/

unfortunately Windows and OSX don't, git-lfs is properly installed and working (I see the 9 MB .mp4 file in the workspace), but the tests don't pass. I'll look into it a bit more, maybe the tests need more work.

wjwwood:
@esteve The OS X job is downloading the file near as I can tell:

From http://ci.ros2.org/job/ci_osx/570/consoleFull:
```
03:12:01 === ./src/ros2/demos (git) ===
03:12:01 Downloading image_tools/test/test_capture.mp4 (9.13 MB)
03:12:01 Switched to a new branch 'image_tools_tests'
03:12:01 Branch image_tools_tests set up to track remote branch image_tools_tests from origin.
```

tfoote:
@wjwwood will look and see if this is still needed.

wjwwood:
@codebot 🍔

PR #56, opened 2015-12-31
don't run the pendulum test with fastrtps

gerkey:
FastRTPS doesn't (yet) support the use of custom allocators, so it will fail this test, which requires that no actual allocation happens during execution.

### DIFF
--- a/pendulum_control/CMakeLists.txt
+++ b/pendulum_control/CMakeLists.txt
@@ -84,39 +84,43 @@ if(AMENT_ENABLE_TESTING)
   set(RCLCPP_DEMO_PENDULUM_TELEOP_EXPECTED_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/test/pendulum_teleop")
 
   foreach(middleware_impl ${middleware_implementations})
-    set(RCLCPP_DEMO_PENDULUM_LOGGER_EXECUTABLE "${CMAKE_CURRENT_BINARY_DIR}${test_executable_subfolder}/pendulum_logger__${middleware_impl}${test_executable_extension}")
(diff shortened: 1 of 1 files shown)


gerkey:
Linux CI (with ros2/realtime_support#28):

http://ci.ros2.org/job/ci_linux/832

pendulum_control/CMakeLists.txt
```diff
@@ -84,39 +84,43 @@ if(AMENT_ENABLE_TESTING)
...
-    ament_add_nose_test(test_demo_pendulum_teleop__${middleware_impl}
-      "${CMAKE_CURRENT_BINARY_DIR}/test_pendulum_teleop__${middleware_impl}.py"
-      TIMEOUT 20)
-    set_tests_properties(test_demo_pendulum_teleop__${middleware_impl}
-    PROPERTIES DEPENDS "test_demo_pendulum_teleop__${middleware_impl} test_demo_pendulum_teleop__${middleware_impl}")
-
+    # FastRTPS doesn't (yet) support the use of custom allocators, so it will
+    # fail this test, which requires that no actual allocation happens during
+    # execution.
+    if(NOT ${middleware_impl} STREQUAL "rmw_fastrtps_cpp")
```
tfoote: This would be a cleaner diff to use an if(rtps) break() rather than indenting the whole block, and could more easily be reverted in the future.

gerkey: I agree, it would look nicer like this:

~~~
    if(${middleware_impl} STREQUAL "rmw_fastrtps_cpp")
      continue()
    endif()
~~~

Unfortunately, `continue()` was [introduced](https://cmake.org/Bug/view.php?id=14013) in CMake 3.2 and we're [targeting](https://github.com/ros2/ros2/wiki/Developer-Guide#cmake) CMake 2.8.

tfoote: That's amazing they didn't add that primative until 3.2. Oh well.

tfoote:
+1

esteve:
+1

PR #57, opened 2016-01-15
build release on Windows

dirk-thomas:
Connect to ament/ament_tools#72


esteve:
+1

esteve:
This might fix the issues we had with OpenCV and also not require installing MSVC for running the demos.

wjwwood:
+1

PR #58, opened 2016-02-05
use generator expressions for configuration specific tests

dirk-thomas:
Debug: http://ci.ros2.org/job/ci_windows/994/
Release: http://ci.ros2.org/job/ci_windows/995/

### DIFF
--- a/intra_process_demo/CMakeLists.txt
+++ b/intra_process_demo/CMakeLists.txt
@@ -97,17 +97,6 @@ if(AMENT_ENABLE_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
-  if(NOT WIN32)
-    set(test_executable_extension "")
-    set(test_executable_subfolder "")
-  else()
-    set(test_executable_extension ".exe")
-    set(test_executable_subfolder "/${CMAKE_BUILD_TYPE}")
-    if("${test_executable_subfolder}" STREQUAL "/")
-      set(test_executable_subfolder "/Release")
-    endif()
-  endif()
-
   # Add each test case.  Multi-executable tests can be specified in
   # semicolon-separated strings, like  exe1:exe2.
   set(demo_tests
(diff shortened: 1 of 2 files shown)


intra_process_demo/CMakeLists.txt
```diff
@@ -127,18 +116,22 @@ if(AMENT_ENABLE_TESTING)
...
         @ONLY
       )
+      file(GENERATE
+        OUTPUT "test_${exe_list_underscore}${target_suffix}_$<CONFIGURATION>.py"
+        INPUT "${CMAKE_CURRENT_BINARY_DIR}/test_${exe_list_underscore}${target_suffix}.py.configured"
+      )
 
       ament_add_nose_test(test_demo_${exe_list_underscore}${target_suffix}
-        "${CMAKE_CURRENT_BINARY_DIR}/test_${exe_list_underscore}${target_suffix}.py"
+        "${CMAKE_CURRENT_BINARY_DIR}/test_${exe_list_underscore}${target_suffix}_$<CONFIGURATION>.py"
```
tfoote: Reading the docs CONFIGURATION is deprecated and CONFIG should be used instead.

dirk-thomas: Can you please point to the docs.

wjwwood: https://cmake.org/cmake/help/v3.0/manual/cmake-generator-expressions.7.html#informational-expressions

tfoote: https://cmake.org/cmake/help/v3.0/manual/cmake-generator-expressions.7.html and theres an odd reference in 2.8.12 https://cmake.org/cmake/help/v2.8.12/cmake.html#section_Generators which is a little ambiguous

dirk-thomas: We are using 2.8.12 and as far as I can see it doesn't have that option: https://cmake.org/cmake/help/v2.8.12/cmake.html#command:add_compile_options

`$<CONFIGURATION>` evaluates to the configuration name.
`$<CONFIG:cfg>` evaluates to a boolean ('1' if config is "cfg", else '0').

tfoote: That's really stupid of them to deprecate a value and move to a new value that had a completely different meaning in their last major version.

tfoote:
+1

wjwwood:
@dirk-thomas I think the +1 covers the connected pr's? Maybe @tfoote disagrees, but I think these are ready for merge.

tfoote:
Yeah, all 3 are good to merge.

PR #59, opened 2016-02-17
fix launch script

dirk-thomas:
Fixes the missing executable bit and generates a separate script for each rmw impl.


jacquelinekay:
+1

PR #60, opened 2016-02-17
use TLSF allocator in demo

jacquelinekay:
A bit of refactoring to use the TLSF allocator in the pendulum demo.

Should have been done last release, forgot because it wasn't ticketed (except in the tutorial page).

needs ros2/realtime_support#30

### DIFF
--- a/pendulum_control/CMakeLists.txt
+++ b/pendulum_control/CMakeLists.txt
@@ -15,6 +15,7 @@ find_package(rclcpp REQUIRED)
 find_package(rmw_implementation_cmake REQUIRED)
 find_package(pendulum_msgs REQUIRED)
 find_package(rttest)
+find_package(tlsf_cpp)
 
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Looks reasonable to me. Does it make sense to run a CI job for that?

jacquelinekay:
yes, for Linux only. The demo has a nose test.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/1015/

jacquelinekay:
I believe the build fails because it needs this ros2/realtime_support#31

I'll make a branch on realtime_support with that name and retrigger.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/1016/ is clean (except for the Python warning which is on master).

tfoote:
+1

PR #62, opened 2016-04-06
use std::isnan

dirk-thomas:
The existing code doesn't compile on Xenial. Now using http://en.cppreference.com/w/cpp/numeric/math/isnan

* http://ci.ros2.org/job/ci_linux/1160/
* http://ci.ros2.org/job/ci_osx/938/
* http://ci.ros2.org/job/ci_windows/1216/


wjwwood:
LGTM, should this go in before the alpha?

PR #63, opened 2016-04-06
Remove unused rmw typesupport call

jacquelinekay:
While working on ros2/rmw#63 I noticed that the result of this call is never used, so I'm deleting it.


dirk-thomas:
+1

PR #67, opened 2016-05-25
Image tools tests redux

wjwwood:
replacement for https://github.com/ros2/demos/pull/48 that does not use a movie file, but generates burgers instead (thanks @codebot).

screen shot 2016-05-24 at 8 25 42 pm


wjwwood:
CI:

- Linux [Build Status](http://ci.ros2.org/job/ci_linux/1342/)
- OS X [Build Status](http://ci.ros2.org/job/ci_osx/1051/)
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/1355/)

wjwwood:
Btw, this new test is disabled for connext, connext dynamic, and fastrtps for various known issues. I put a comment in the cmake which disables it.

wjwwood:
Ok, it's broken on Windows and OS X. OS X is a problem with how OpenCV is linked in Homebrew and I don't know how to fix it generically. Windows is a mystery too. It's got something to do with how I'm using `file(GENERATE ...)`. I'm going to put this on hold until after the alpha. I've wasted way too much time on it. If someone else wants to pick it up before then, that'd be great.

wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1716/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1282/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1638/)

image_tools/CMakeLists.txt
```diff
@@ -50,6 +53,56 @@ call_for_each_rmw_implementation(targets GENERATE_DEFAULT)
...
+      "${rmw_implementation} " STREQUAL "rmw_connext_dynamic_cpp " OR
+      # TODO(wjwwood): fastrtps fails with:
+      #
+      #   uncaught exception of type std::runtime_error: vector overcomes the maximum length
+      #
+      # We can remove this disable for fastrtps when that's fixed.
+      "${rmw_implementation} " STREQUAL "rmw_fastrtps_cpp "
+    )
+      message(STATUS "Skipping tests for '${rmw_implementation}'")
+      return()
```
dirk-thomas: I assume the purpose of this return is to skip the current rmw implementation. But `return` doesn't work like that in a `macro`. Please see https://cmake.org/cmake/help/v3.5/command/return.html

dirk-thomas:
It also looks like this change broke the Windows build when using OpenSplice: see http://ci.ros2.org/job/ci_windows/1650/console

wjwwood:
I'll look at it, but the CI jobs don't test OpenSplice any longer.

image_tools/CMakeLists.txt
wjwwood: This should be resolved in https://github.com/ros2/demos/pull/78.

dirk-thomas:
I created a ticket for it: #79

PR #70, opened 2016-06-01
fix polluting source space

dirk-thomas:
Fixes #65.

* [Build Status](http://ci.ros2.org/job/ci_linux/1371/)
* [Build Status](http://ci.ros2.org/job/ci_osx/1068/)
* [Build Status](http://ci.ros2.org/job/ci_windows/1377/)

Connect to #65.

PR #71, opened 2016-06-03
Fix flaky test in pendulum demo

dhood:
The output regex was a bit too strict.. not a common point of failure but I have on at least one occasion gotten:
```
5: [test_executable_1] Latency (time after deadline was missed):
5: [test_executable_1] - Min: 6774 ns
5: [test_executable_1] - Max: 2951825 ns
5: [test_executable_1] - Mean: 65343.5 ns
5: [test_executable_1] - Standard deviation: 1370
```
instead of the more common standard deviation format of 1367.81 or 1.35819e+08, for example.

http://ci.ros2.org/job/ci_linux/1386/ has finished without issues but I've triggered another round since #70 was merged:

[Build Status](http://ci.ros2.org/job/ci_linux/1387/)
[Build Status](http://ci.ros2.org/job/ci_osx/1084/)
[Build Status](http://ci.ros2.org/job/ci_windows/1398/)

### DIFF
--- a/pendulum_control/test/pendulum_demo.regex
+++ b/pendulum_control/test/pendulum_demo.regex
@@ -7,7 +7,7 @@ rttest statistics:
     - Min: \d+ ns
     - Max: \d+ ns
     - Mean: \d+.\d+ ns
-    - Standard deviation: \d+\.\d+(e\+\d+)?
+    - Standard deviation: \d+(\.\d+)?(e[\+\-]\d+)?
 
 
 PendulumMotor received \d+ messages


dirk-thomas:
Nice +1

dhood:
the `test_pendulum__rmw_*_cpp` tests passed for the linux CI jobs 1386 and 1387 (they don't target other platforms). stddev outputs of types 1298.19 and 1.35819e+08 were printed in those jobs. I admit that there hasn't been an output that stresses the case of e.g. 1370 but I have tested the regex elsewhere

wjwwood:
CI report is ok, and there's been no new changes since @dirk-thomas's +1, so I'm going to merge this.

PR #74, opened 2016-07-01
Prevent pendulum_control demos from using typesupport introspection

dhood:
[This commit](https://github.com/ros2/rosidl/pull/158/commits/1e4b0741cbd82da5e62b0822ab24f293ede12ba4#diff-4845b4409f48ba8429ddbfdb2d93816fR8) caused pendulum_control tests to start failing for rmw implementations without typesupport introspection support e.g. http://ci.ros2.org/job/ci_linux/1615/testReport/(root)/projectroot/test_demo_pendulum_teleop__rmw_connext_cpp/

This fixes the failing `pendulum_control` tests for connext static (and opensplice). 

http://ci.ros2.org/job/ci_linux/1627/testReport/ (pendulum demos don't run on windows or os x)

(after approved I'll fix the commit history)

### DIFF
--- a/pendulum_control/CMakeLists.txt
+++ b/pendulum_control/CMakeLists.txt
@@ -32,23 +32,23 @@ macro(targets)
   add_executable(pendulum_demo${target_suffix}
     src/pendulum_demo.cpp)
   ament_target_dependencies(pendulum_demo${target_suffix}
-    "pendulum_msgs"
     "rclcpp${target_suffix}"
+    "pendulum_msgs"
     "rttest"
     "tlsf_cpp")
 
   add_executable(pendulum_logger${target_suffix}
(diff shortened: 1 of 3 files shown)


dirk-thomas:
lgtm

wjwwood:
lgtm too, make sure you squash on commit, it looks like you have a merge commit and possibly a duplicate commit in here.

dhood:
have removed them now. won't squash since they're unrelated commits

wjwwood:
sounds good

mikaelarguedas:
you may want to add a comment explaining why the dependency order is not alphabetical. This would prevent someone to reintroduce the failure in a further "cleanup/linting" commit

dhood:
that's a good point - will do that.

PR #76, opened 2016-07-11
add conversion string for 32-bit floating-point images to showimage

codebot:
so it can render depth images from the Astra


gerkey:
+1. I saw it work in person.

wjwwood:
Lgtm

PR #77, opened 2016-08-26
fix missing header

Harin55:
The use of `M_PI` requires the standard library math.h


wjwwood:
Thanks, it looked good to me, so I merged it. However, I also updated it in a subsequent commit to use `<cmath>` since that's the cpp idiomatic thing to do.

Thanks again for the contribution!

PR #80, opened 2016-10-01
comply with stricter cpplint rules

dirk-thomas:
Update style to pass new cpplint check (ament/ament_lint#59).

http://ci.ros2.org/job/ci_linux/1763/

### DIFF
--- a/intra_process_demo/src/image_pipeline/camera_node.cpp
+++ b/intra_process_demo/src/image_pipeline/camera_node.cpp
@@ -16,6 +16,8 @@
 
 #include <rclcpp/rclcpp.hpp>
 
+#include <memory>
+
 int main(int argc, char ** argv)
 {
   rclcpp::init(argc, argv);
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <iostream>
+#include <memory>
 #include <string>
 
 #include "opencv2/highgui/highgui.hpp"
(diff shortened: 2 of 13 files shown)


intra_process_demo/src/image_pipeline/camera_node.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include <rclcpp/rclcpp.hpp>
```
Karsten1987: Can we be consistent on the order and how to include the rcl headers? see [this demo](https://github.com/ros2/demos/blob/master/intra_process_demo/src/cyclic_pipeline/cyclic_pipeline.cpp#L20) for comparison

dirk-thomas: I agree that having `memory` further up in the list is preferred. But in some cases (e.g. https://github.com/ros2/demos/pull/80/files#diff-c53d060e7c287a653e8fef96c7f7a4eeR33) that would require quite a few of includes to be changed from `<...>` to `"..."`. Not sure if we want to do that?

wjwwood: I'd vote to merge as is and ticket the annoyance. For the long term solution I'd like to see cpplint.py patched to allow something like this:

```c++
#include <memory>

#include <rclcpp/rclcpp.hpp>

#include "foo/foo.hpp"
```

dirk-thomas: Done: ament/ament_lint#62

Issue #68, opened 2016-06-01
demos should print reasonable error message when no camera is present


dirk-thomas:
Currently `camera_node` and `image_pipeline_*` just crash without any notice (on Windows). `cam2image` e.g. prints an error message (`Could not open video stream`) which is the desired behavior.

wjwwood:
We can add error handling to all of the demos (`try {...} catch(...) {...}`) or we could try to put something automatic together with some combination of these c++11 functions:

- http://en.cppreference.com/w/cpp/error/set_terminate
- http://en.cppreference.com/w/cpp/error/current_exception
- http://en.cppreference.com/w/cpp/error/uncaught_exception

jacquelinekay:
I would guess from this that Windows doesn't print the message for an exception when it is thrown by default? That's annoying.

+1 for checking for uncaught_exception in the destructor of CameraNode, rather than wrapping the instantiation of CameraNode with try/catch everywhere.

dirk-thomas:
@jacquelinekay The exception is raised in https://github.com/ros2/demos/blob/82bbf8fefafcfed5f9bf8e844821ab81b564f1bf/intra_process_demo/src/image_pipeline/camera_node.hpp#L43 which makes the object creation fail. There is no invocation of the destructor anywhere.

wjwwood:
We could have a standard "termination_handler" that uses "current_exception" to print a useful message, and then we'd only need to install it with "set_terminate" in each demo. That's slightly less intrusive than wrapping all mains with a try-catch, but also slightly more magical.

jacquelinekay:
hmm yeah, I was misinterpreting the example on http://en.cppreference.com/w/cpp/error/uncaught_exception

Karsten1987:
any news on that? I believe the actual error comes from https://github.com/ros2/demos/blob/master/intra_process_demo/src/image_pipeline/camera_node.hpp#L39 where it blindly opens device 0

Karsten1987:
```
c:\dev\ci-ros2-fastrtps\ros2-windows>camera_node
Could not open video stream! Exiting..
```

fixed in https://github.com/ros2/demos/pull/81

PR #81, opened 2016-10-04
(fix) quickfix for printing error message on windows

Karsten1987:
not the most elegant solution, but it works and correctly prints the error message on windows. The tutorials are complete and do make sense to the users with this.

see https://github.com/ros2/demos/issues/68

### DIFF
--- a/intra_process_demo/src/image_pipeline/camera_node.cpp
+++ b/intra_process_demo/src/image_pipeline/camera_node.cpp
@@ -21,7 +21,16 @@
 int main(int argc, char ** argv)
 {
   rclcpp::init(argc, argv);
-  auto camera_node = std::make_shared<CameraNode>("image");
-  rclcpp::spin(camera_node);
-  return 0;
+  try
+  {
+    auto camera_node = std::make_shared<CameraNode>("image");
+    rclcpp::spin(camera_node);
+    return 0;
+  }
+  catch( const std::exception& e )
+  {
+    fprintf(stderr, "Failed to open camera node. Exiting..\n");
+    return 1;
+  }
+  
 }
--- a/intra_process_demo/src/image_pipeline/camera_node.cpp
+++ b/intra_process_demo/src/image_pipeline/camera_node.cpp
@@ -21,16 +21,15 @@
 int main(int argc, char ** argv)
 {
   rclcpp::init(argc, argv);
-  try
-  {
-    auto camera_node = std::make_shared<CameraNode>("image");
-    rclcpp::spin(camera_node);
-    return 0;
-  }
-  catch( const std::exception& e )
-  {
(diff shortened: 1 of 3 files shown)


intra_process_demo/src/image_pipeline/camera_node.cpp
```diff
@@ -21,7 +21,16 @@
...
 {
   rclcpp::init(argc, argv);
-  auto camera_node = std::make_shared<CameraNode>("image");
-  rclcpp::spin(camera_node);
-  return 0;
+  try
+  {
+    auto camera_node = std::make_shared<CameraNode>("image");
+    rclcpp::spin(camera_node);
+    return 0;
```
wjwwood: Only the first line needs to be in the try-catch.

intra_process_demo/src/image_pipeline/camera_node.cpp
```diff
@@ -21,7 +21,16 @@
...
+  }
+  catch( const std::exception& e )
```
wjwwood: I think you need to run `ament_uncrustify` over these changes, I would expect the style to be:

```c++
} catch(const std::exception & e) {
```

intra_process_demo/src/image_pipeline/camera_node.cpp
```diff
@@ -21,7 +21,16 @@
...
+  {
+    fprintf(stderr, "Failed to open camera node. Exiting..\n");
```
wjwwood: So long as you're catching the exception, you might as well include the `.what()` of the exception in the message. Otherwise, I'd suggest catching the exact error that is being thrown.

intra_process_demo/src/image_pipeline/camera_node.cpp
Karsten1987: is there any best practice of keeping the try-catch as short as possible over forward declaring the pointer and then proof for `nullptr`?

wjwwood:
@dirk-thomas should we wait for after the alpha to merge this or try to get it in?

intra_process_demo/src/image_pipeline/image_pipeline_all_in_one.cpp
```diff
@@ -26,7 +26,13 @@ int main(int argc, char * argv[])
   rclcpp::executors::SingleThreadedExecutor executor;
 
   // Connect the nodes as a pipeline: camera_node -> watermark_node -> image_view_node
-  auto camera_node = std::make_shared<CameraNode>("image");
+  std::shared_ptr<CameraNode> camera_node = nullptr;
+  try {
+    auto camera_node = std::make_shared<CameraNode>("image");
```
dirk-thomas: The variable get redeclared in this line.

PR #82, opened 2016-10-26
Consistent naming when using CMake variable for rmw implementation

dhood:
The pendulum demo_teleop test was failing in http://ci.ros2.org/job/ci_linux/1819/ because `rmw_implementation` was being templated into [the test script](https://github.com/ros2/demos/blob/700e32d0c55df515c83a59c03a33dce321299abb/pendulum_control/test/test_pendulum_teleop.py.in) instead of `middleware_impl`.

I've unified these two variables so that the correct RMW implementation's output is filtered, and furthermore so mistakes like this don't happen again. I've extended the change to other packages too that were using the two variables.

Linux [Build Status](http://ci.ros2.org/job/ci_linux/1823/)
OS X [Build Status](http://ci.ros2.org/job/ci_osx/1359/)
Windows [Build Status](http://ci.ros2.org/job/ci_windows/1732/)

### DIFF
--- a/image_tools/CMakeLists.txt
+++ b/image_tools/CMakeLists.txt
@@ -55,9 +55,9 @@ if(BUILD_TESTING)
   ament_lint_auto_find_test_dependencies()
 
   # get the rmw implementations ahead of time
-  get_available_rmw_implementations(middleware_implementations)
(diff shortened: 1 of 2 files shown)


dhood:
jobs back to passing without test failures - merging.

PR #83, opened 2016-10-26
Enable image_tools tests with synthetic image

dhood:
connects to ros2/demos#43
todo: test different quality of service configurations

### DIFF
--- a/image_tools/CMakeLists.txt
+++ b/image_tools/CMakeLists.txt
@@ -75,7 +75,7 @@ if(BUILD_TESTING)
       ENV
       RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation}
       RMW_IMPLEMENTATION=${rmw_implementation}
-      TIMEOUT 60
+      TIMEOUT 30
     )
   endmacro()
 
--- a/image_tools/CMakeLists.txt
+++ b/image_tools/CMakeLists.txt
@@ -45,33 +45,18 @@ if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
+  # get the rmw implementations ahead of time
   find_package(rmw_implementation_cmake REQUIRED)
+  get_available_rmw_implementations(rmw_implementations)
+  foreach(rmw_implementation ${rmw_implementations})
+    find_package("${rmw_implementation}" REQUIRED)
+  endforeach()
 
   # These are the regex's for validating the output of the executables.
   set(RCLCPP_DEMO_SHOWIMAGE_EXPECTED_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/test/showimage")
   set(RCLCPP_DEMO_CAM2IMAGE_EXPECTED_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/test/cam2image")
 
   macro(testing_targets)
-    set(SKIP_TEST "")
-    if(
-      # TODO(wjwwood): connext and connext_dynamic fail with:
-      #
-      #   COMMENDSrWriterService_write:!write. Reliable large data requires asynchronous writer.
-      #
-      # We can remove this disable for connext when that's fixed.
-      "${rmw_implementation} " STREQUAL "rmw_connext_cpp " OR
-      "${rmw_implementation} " STREQUAL "rmw_connext_dynamic_cpp " OR
-      # TODO(wjwwood): fastrtps fails with:
-      #
-      #   uncaught exception of type std::runtime_error: vector overcomes the maximum length
-      #
-      # We can remove this disable for fastrtps when that's fixed.
-      "${rmw_implementation} " STREQUAL "rmw_fastrtps_cpp "
-    )
-      message(STATUS "Skipping tests for '${rmw_implementation}'")
-      set(SKIP_TEST "SKIP_TEST")
-    endif()
-
     set(RCLCPP_DEMO_CAM2IMAGE_EXECUTABLE $<TARGET_FILE:cam2image>)
     set(RCLCPP_DEMO_SHOWIMAGE_EXECUTABLE $<TARGET_FILE:showimage>)
 
@@ -80,15 +65,17 @@ if(BUILD_TESTING)
       test_showimage_cam2image${target_suffix}.py.genexp
       @ONLY
     )
+
     file(GENERATE
       OUTPUT test_showimage_cam2image${target_suffix}_$<CONFIG>.py
       INPUT test_showimage_cam2image${target_suffix}.py.genexp)
 
     ament_add_nose_test(test_showimage_cam2image${target_suffix}
       "${CMAKE_CURRENT_BINARY_DIR}/test_showimage_cam2image${target_suffix}_$<CONFIG>.py"
-      ENV RMW_IMPLEMENTATION=${rmw_implementation}
+      ENV
+      RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation}
+      RMW_IMPLEMENTATION=${rmw_implementation}
       TIMEOUT 60
-      ${SKIP_TEST}
     )
   endmacro()
 
--- a/image_tools/CMakeLists.txt
+++ b/image_tools/CMakeLists.txt
@@ -45,12 +45,7 @@ if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
-  # get the rmw implementations ahead of time
   find_package(rmw_implementation_cmake REQUIRED)
-  get_available_rmw_implementations(rmw_implementations)
-  foreach(rmw_implementation ${rmw_implementations})
-    find_package("${rmw_implementation}" REQUIRED)
-  endforeach()
 
   # These are the regex's for validating the output of the executables.
   set(RCLCPP_DEMO_SHOWIMAGE_EXPECTED_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/test/showimage")
--- a/image_tools/test/cam2image.txt
+++ b/image_tools/test/cam2image.txt
@@ -1,0 +1,1 @@
+Publishing image #1
--- a/image_tools/test/cam2image.txt
+++ b/image_tools/test/cam2image.txt
@@ -1,1 +1,2 @@
+Publishing data on topic 'image'
 Publishing image #1
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstdio>
 #include <iostream>
 #include <memory>
 #include <string>
@@ -84,6 +85,11 @@ int main(int argc, char * argv[])
   size_t height = 240;
   bool burger_mode = false;
   std::string topic("image");
+
+  // Force flush of the stdout buffer.
+  // This ensures a correct sync of all prints
+  // even when executed simultaneously within a launch file.
+  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
 
   // Configure demo parameters with command line options.
   bool success = parse_command_options(
--- a/image_tools/src/showimage.cpp
+++ b/image_tools/src/showimage.cpp
@@ -92,6 +92,11 @@ int main(int argc, char * argv[])
   bool show_camera = true;
   std::string topic("image");
 
+  // Force flush of the stdout buffer.
+  // This ensures a correct sync of all prints
+  // even when executed simultaneously within a launch file.
+  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
+
   // Configure demo parameters with command line options.
   if (!parse_command_options(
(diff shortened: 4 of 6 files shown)


dhood:
update: different reliability QoS settings now tested.

todo: tests not running properly on windows http://ci.ros2.org/job/ci_windows/1740/testReport/(root)/test_showimage_cam2image__rmw_fastrtps_cpp_Release/test_reliable_qos/

dhood:
Update: 
1. runs on windows now. all that was needed was to add the opencv bin directory to the path of the buildfarm machines.
1. there's still some bad performance with fastrtps.


it takes a long time for an image to be received when using best effort publishing and subscribing with FastRTPS. [see this test output](http://ci.ros2.org/job/ci_windows/2254/testReport/(root)/test_showimage_cam2image__rmw_fastrtps_cpp_Release/test_best_effort_qos/). This does not happen for reliable pub+sub. Interestingly, this also does not happen for a reliable publisher and a best effort subscriber, where the resulting connection should be best effort.

it seems that the best effort publisher might be what's causing the performance problems. there might be something in the reliable publisher that makes it easier for the subscriber (best effort or reliable) to get the message. I am not sure why at the moment.

If others agree I will focus on getting this merged just to confirm that these demos are runnable, and we can investigate the behaviour issues as a separate task. I'll reduce the publishing rate of the images so that the timeout doesn't have to be so high to accommodate fastrtps (it's 60s).

dhood:
another update: only reliable is working for os x. best effort publisher and/or subscriber does not show that any message is received. I don't remember this being an issue when I first started working on this PR.

The changes referenced in https://github.com/ros2/rmw_fastrtps/issues/81#issuecomment-269345234 might be responsible (I don't know that we have any other tests for best effort communication + fragmentation, so that might be why we didn't notice this before.)

mikaelarguedas:
That sounds good to me. +1 for getting this merged without fixing the fast-RTPS throughput issue in best effort

Rationale:
The best effort throughput issue on FastRTPS may be out of the scope of this PR and I think that several issues and discussion are reporting the same problem. Edit: like the on referenced in the previous comment

>  (I don't know that we have any other tests for best effort communication + fragmentation, so that might be why we didn't notice this before.)

+1 I think this is the only demo actually testing best effort on big messages. To my knowledge the only other tests with big messages are in `test_communication` where we established only reliable communication

dhood:
So, the reason it wasn't working on osx with best effort is because of packet collisions. enabling the throughput controller for fastrtps means that the fragments of the image aren't all sent at the same time and so the collisions are less likely.

With a throughput controller of 30,000 bytes/10ms and a heartbeat period of 10ms, all platforms and all reliability combinations can receive an image using the default settings of 320x240 @ 30fps in less than a second:
http://ci.ros2.org/job/ci_osx/1740/testReport/(root)/test_showimage_cam2image__rmw_fastrtps_cpp_/
http://ci.ros2.org/job/ci_windows/2266/testReport/(root)/test_showimage_cam2image__rmw_fastrtps_cpp_Release/
http://ci.ros2.org/job/ci_linux/2267/testReport/(root)/test_showimage_cam2image__rmw_fastrtps_cpp_/

I have opened https://github.com/ros2/rmw_fastrtps/issues/92 to track the discussion on what to do with this moving forward. For now, now that the reason these tests were having different behaviour on different platforms is clearer, I will only enable the reliable tests for this demo.

dhood:
> now that the reason these tests were having different behaviour on different platforms is clearer

As @mikaelarguedas pointed out this isn't 100% accurate - it's still not clear _why_ the behaviour is dependent on the operating system, just that the throughput controller is a fix.

To be explicit, without any changes to fastrtps (ie not throughput controller, default heartbeat period):
Performance is [fine on linux](http://ci.ros2.org/job/ci_linux/2254/testReport/(root)/test_showimage_cam2image__rmw_fastrtps_cpp_/): 0.6s to receive a best effort image
Performance is [poor on Windows](http://ci.ros2.org/job/ci_windows/2254/testReport/(root)/test_showimage_cam2image__rmw_fastrtps_cpp_Release/): 28s to receive a best effort image
Performance is [bad on OS X](http://ci.ros2.org/job/ci_osx/1707/testReport/(root)/projectroot/test_demo_showimage_cam2image__rmw_fastrtps_cpp/): no best effort image received in 60s

---

CI for just this branch, not dependent on any changes to fastrtps:
windows: [Build Status](http://ci.ros2.org/job/ci_windows/2271/)
os x: [Build Status](http://ci.ros2.org/job/ci_osx/1744/) (existing test failure)
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2270/)

wjwwood:
How are you measuring that? Just the time the tests take to complete?

dhood:
yes, with the test completion time. the tests will pass once _any_ image has been received.

mikaelarguedas:
Yeah still wondering how there can be collision with a single sender o_O. maybe a magic ability to send simultaneously unicast and multicast packets :unicorn: that end up colliding...

I think you'll want to rebase this before merging (or use the "rebase and merge" option) otherwise the tests marked as _skipped_ won't be skipped anymore.

dhood:
I have edited my comment to clarify that I was linking to two different sets of CI jobs: one with best effort tests enabled (top), and one with them not enabled (bottom), the current state of this PR.

> maybe a magic ability to send simultaneously unicast and multicast packets :unicorn: that end up colliding...

Yeah I was just speaking with @wjwwood about that. I do see that, at least on my linux machine, I can see unicast and multicast packets being sent.

> I think you'll want to rebase this before merging (or use the "rebase and merge" option) otherwise the tests marked as skipped won't be skipped anymore.
I am not sure that I understand. I do not want the "showimage_cam2image" tests to be skipped anymore. Are you speaking about another?

wjwwood:
It could also be that the kernel's udp buffer is getting overwhelmed by the bursty nature without the flow controller, but that's just conjecture.

What's the evidence for the package collision as described above? I didn't see anything in @dhood's post where she mentions that.

dhood:
> What's the evidence for the package collision as described above? I didn't see anything in @dhood's post where she mentions that.

I don't have evidence that it's packet collision - I shouldn't have jumped to conclusions on that one.

As an additional data point, I observed that running the demo on OS X at higher frequency (e.g. 100 Hz instead of 30 Hz) allows messages to be received by the best effort subscriber, even without any throughput controller....

mikaelarguedas:
>> I think you'll want to rebase this before merging (or use the "rebase and merge" option) otherwise the tests marked as skipped won't be skipped anymore.

> I am not sure that I understand. I do not want the "showimage_cam2image" tests to be skipped anymore. Are you speaking about another?

Sorry I should have phrased it differently, I just wanted to be sure you were aware that this was reenabling all of them. Haven't tried any `connext_dynamic` code for a while so I just assumed it was still broken.

wjwwood:
> As an additional data point, I observed that running the demo on OS X at higher frequency (e.g. 100 Hz instead of 30 Hz) allows messages to be received by the best effort subscriber, even without any throughput controller....

lol wut

dhood:
> Haven't tried any connext_dynamic code for a while so I just assumed it was still broken.

I think that the decision we made once upon a time was that we only disable tests for "on pause" rmw implementations e.g. https://github.com/ros2/system_tests/blob/3005b6df5261098bc6a51567b57f1022f7ef8830/test_communication/CMakeLists.txt#L60 if they are being skipped because of known implementation-specific bugs. I don't think we skip tests for them otherwise. so enabling for all rmw implementations is OK as I see it

dhood:
> Haven't tried any connext_dynamic code for a while so I just assumed it was still broken.

OK you're referring to the "COMMENDSrWriterService_write:!write. Reliable large data requires asynchronous writer." comment that's in the code. We use asynchronous writer now [in the shared connext functions](https://github.com/ros2/rmw_connext/blob/2f7630d17deee3cdc3e7e16d8c538d62afff832e/rmw_connext_shared_cpp/src/shared_functions.cpp#L261), so it should be fine to re-enable this even for connext dynamic.

mikaelarguedas:
sounds good to me :+1:  thanks for clarifying

dhood:
> > As an additional data point, I observed that running the demo on OS X at higher frequency (e.g. 100 Hz instead of 30 Hz) allows messages to be received by the best effort subscriber, even without any throughput controller....

> lol wut

I've checked on the three OS X buildfarm machines and they all had the same thing.. wouldn't receive images at 30 Hz, would receive images at 100 Hz (eyeballing it it seems to actually be at 100 Hz, too). It evens improves further if you increase the frequency more....

> Yeah still wondering how there can be collision with a single sender o_O. maybe a magic ability to send simultaneously unicast and multicast packets :unicorn: that end up colliding...

Implementing the changes [suggested here](https://github.com/ros2/rmw_fastrtps/issues/36#issuecomment-228753919) to only use multicast for subscribers seems to improve performance a bit (it can receive an image every so often, varies with runs), but there still seems to be something preventing images from being received in whole.

I've opened https://github.com/ros2/rmw_fastrtps/issues/93 to ask the eProsima folks if they have any ideas.

dhood:
I didn't realise that this was labelled as in review. Could I get an explicit approval from someone? As we discussed above, this PR is just to check that the executables are runnable and while it was useful for evaluating fastrtps' performance, its scope has been narrowed to not depend on any changes to rmw_fastrtps.

image_tools/CMakeLists.txt
```diff
@@ -45,33 +45,18 @@ if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
 
+  # get the rmw implementations ahead of time
   find_package(rmw_implementation_cmake REQUIRED)
+  get_available_rmw_implementations(rmw_implementations)
+  foreach(rmw_implementation ${rmw_implementations})
+    find_package("${rmw_implementation}" REQUIRED)
+  endforeach()
```
dirk-thomas: Is this really necessary?

dhood: The reason that I thought it was necessary is because it seems from [your comment here](https://github.com/ros2/ros2/issues/302#issuecomment-276519274) that these lines [were not removed from the tests mentioned](https://github.com/ros2/system_tests/blob/3005b6df5261098bc6a51567b57f1022f7ef8830/test_rclcpp/CMakeLists.txt#L33..L38) for a reason that also applies here (we are making test executables for each rmw implementation).

Will the `call_for_each_rmw_implementation` macro work fine without doing `find_package` on each of the rmw implementations? I thought it must be necessary but since [that macro calls `get_all_rmw_implementations`](https://github.com/ros2/rmw/blob/847784e6a0b714630c12a67eb753439ae234121a/rmw_implementation_cmake/cmake/call_for_each_rmw_implementation.cmake#L49) itself perhaps this block can be removed again.

dirk-thomas: I don't see where this is true (if we consider "executable" to be a compiled executable rather than an executable Pythonscript):

> we are making test executables for each rmw implementation

As far as I can see the `testing_targets` macros doesn't contain any compiled targets. It only expands a Python template and registers that Python file as a nosetest. So there is nothing which uses the `find_package` information. For the referenced tests in `system_tests` that is different - the function there actually builds executables for each rmw impl.

dhood: OK I see now that the only info that's needed is the list of RMW implementations from `get_all_rmw_implementations`, not the packages themselves. I will remove this block thanks.

dhood:
Addressed @dirk-thomas' comment in https://github.com/ros2/demos/pull/83/commits/11cd2e3337a06242f0f0293d81ab13942a5f2a14

Rebased and adapted for https://github.com/ros2/demos/pull/117 in https://github.com/ros2/demos/pull/83/commits/86c3d36120b1463bb40b442c3e343346afee9761

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/2309/)
OS X: [Build Status](http://ci.ros2.org/job/ci_osx/1757/) (standard test failure these days)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2286/)

image_tools/test/cam2image.txt
```diff
@@ -0,0 +1,2 @@
+Publishing data on topic 'image'
+Publishing image #1
```
mikaelarguedas: Can you give more context about why renaming this file was necessary ? Asking because I thought that launch testing was explicitely looking for `.regex` files

mikaelarguedas: Sorry missing a pointer here: based on [this block](https://github.com/ros2/launch/blob/master/launch_testing/launch_testing/__init__.py#L132-L139) I would expect this `cam2image.txt` file to never be taken into account  during the tests

dhood: launch testing can work with either `.txt` or `.regex`. `.txt` has to match explicitly. `cam2image.txt` will be matched as a literal file because of the particular block that you referenced

mikaelarguedas: oops messed up my search for `.txt` :confused: sorry for the noise

dhood:
I've run CI again on this since it's been a few weeks and I'm going to merge this.
linux [Build Status](http://ci.ros2.org/job/ci_linux/2354/) (linters on another package)
windows [Build Status](http://ci.ros2.org/job/ci_windows/2457/)
osx [Build Status](http://ci.ros2.org/job/ci_osx/1808/)

PR #86, opened 2016-11-29
(refactor) add lifecycle demo

Karsten1987:
demo applications for lifecycle (C++)

Connects to ros2/rclcpp#265


Karsten1987:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1562/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/2022/)

Windows seems to be a unrelated flaky test

lifecycle/src/lifecycle_listener.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include <memory>
+#include <string>
+
+#include "lifecycle_msgs/msg/transition_event.hpp"
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+static constexpr auto chatter_topic = "lifecycle_chatter";
+static constexpr auto notification_topic = "lc_talker__transition_event";
```
wjwwood: Since these two strings are only used in one place each, you could simplify this and use them in place.

lifecycle/src/lifecycle_service_client.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include "lifecycle_msgs/msg/transition.hpp"
+#include "lifecycle_msgs/srv/change_state.hpp"
+#include "lifecycle_msgs/srv/get_state.hpp"
+
+#include "rclcpp/rclcpp.hpp"
+
+// service topics for node-attached services
+static constexpr char const * lifecycle_node = "lc_talker";
+static constexpr char const * node_get_state_topic = "lc_talker__get_state";
+static constexpr char const * node_change_state_topic = "lc_talker__change_state";
```
wjwwood: These last two are only used in one other place (`lifecycle_node` is used in several places), so you could remove the last two and use them in place to simplify this demo.

lifecycle/src/lifecycle_service_client.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  unsigned int
+  get_state(std::chrono::seconds time_out = 3_s)
+  {
+    auto request = std::make_shared<lifecycle_msgs::srv::GetState::Request>();
+
+    if (!client_get_state_->wait_for_service(time_out)) {
+      fprintf(stderr, "Service %s is not available.\n",
+        client_get_state_->get_service_name().c_str());
+      return lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN;
```
wjwwood: Should this not throw instead?

lifecycle/src/lifecycle_service_client.cpp
```diff
@@ -0,0 +1,161 @@
...
+    request->transition.id = transition;
+
+    if (!client_change_state_->wait_for_service(time_out)) {
+      fprintf(stderr, "Service %s is not available.\n",
+        client_change_state_->get_service_name().c_str());
+      return false;
+    }
+
+    auto result = client_change_state_->async_send_request(request);
+    if (result.wait_for(std::chrono::milliseconds(time_out)) != std::future_status::ready) {
```
wjwwood: This is tricky, since it only works because you are spinning in a different thread. It might warrant a comment.

lifecycle/src/lifecycle_service_client.cpp
```diff
@@ -0,0 +1,161 @@
...
+  std::shared_ptr<rclcpp::client::Client<lifecycle_msgs::srv::ChangeState>> client_change_state_;
+};
+
+void
+callee_script(std::shared_ptr<LifecycleServiceClient> lc_client)
+{
+  auto sleep_time = 10_s;
+
+  {  // configure
+    lc_client->change_state(lifecycle_msgs::msg::Transition::TRANSITION_CONFIGURE);
```
wjwwood: Should you be checking the returned value of this function to ensure it changed states?

lifecycle/src/lifecycle_talker.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+private:
+  std::shared_ptr<std_msgs::msg::String> msg_;
+  std::shared_ptr<rclcpp_lifecycle::LifecyclePublisher<std_msgs::msg::String>> pub_;
+  std::shared_ptr<rclcpp::TimerBase> timer_;
+};
+
+int main(int argc, char * argv[])
+{
+  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
```
wjwwood: What is this for?

lifecycle/src/lifecycle_service_client.cpp
Karsten1987: This is actually responsible for Ctrl-C not working correctly. Do we have a "best-practice" for a blocking service call?

lifecycle/src/lifecycle_talker.cpp
Karsten1987: i copied that from existing demos. Claiming it as being responsible for correct prints on windows.

lifecycle/src/lifecycle_service_client.cpp
wjwwood: No we don't, because we do not have a way to block on a service call while spin is happening in a different thread or from within a spin thread.

You could modify this to check `rclcpp::ok()` periodically:

```c++
auto end = std::chrono::steady_clock::now() + std::chrono::milliseconds(time_out);
std::chrono::milliseconds wait_period(100);
std::future_status status;
do {
  auto now = std::chrono::steady_clock::now();
  auto time_left = end - now;
  if (time_left <= 0) {break;}
  status = result.wait_for((time_left < wait_period) ? time_left : wait_period);
} while (rclcpp::ok() && status != std::future_status::ready)

if (status != std::future_status::ready) {
  // ... error handling
}
```

It's a bit tedious, but it accomplishes the goal. You could imagine generalizing this into a function in rclcpp, which a signature like this:

```c++
// throw exception when rclcpp::ok() != true
template<typename FutureT, typename WaitTimeT, typename WaitPeriodT = std::chrono::milliseconds>
std::future_status
rclcpp::wait_for_future(
  FutureT & future,
  WaitTimeT time_to_wait,
  WaitPeriodT max_wait_period = std::chrono::milliseconds(100));
```

lifecycle/src/lifecycle_service_client.cpp
```diff
@@ -0,0 +1,295 @@
...
+  auto end = std::chrono::steady_clock::now() + time_to_wait;
+  std::chrono::milliseconds wait_period(100);
+  std::future_status status;
+  do {
+    auto now = std::chrono::steady_clock::now();
+    auto time_left = end - now;
+    if (time_left <= std::chrono::seconds(0)) {break;}
+    status = future.wait_for((time_left < wait_period) ? time_left : wait_period);
+  } while (rclcpp::ok() && status != std::future_status::ready);
+  return status;
```
dirk-thomas: This results in a compiler warning on at least Linux:

> ‘status’ may be used uninitialized in this function [-Wmaybe-uninitialized]

dirk-thomas: Fixed by #92.

PR #90, opened 2016-12-12
QoS demo - topic monitor

dhood:
This includes a `topic_monitor` that's useful for showing QoS settings' behaviours visually and some launch files for comparing different QoS profiles.

The topic monitor subscribes to topics that match a particular pattern, expects periodic messages with increasing values, and calculates the reception rate accordingly.

It can either plot the reception rate itself with matplotlib (useful for windows), or it publishes them to a topic that can be bridged and plotted with rqt.

By default it subscribes as reliable but it will subscribe as best effort if "_best_effort" is at the end of the topic name.


topic_monitor/scripts/data_publisher.py
```diff
@@ -0,0 +1,136 @@
...
+        '--transient-local',
+        action='store_true',
+        default=False,
+        help='Set QoS durability option to "transient local"')
+
+    parser.add_argument(
+        '--depth',
+        type=int,
+        default=default_depth,
+        action='store',
```
mikaelarguedas: Isn't store the default behavior of argparse? if so all `action='store'` can be removed, in all files

topic_monitor/scripts/data_publisher.py
```diff
@@ -0,0 +1,136 @@
...
+        else QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABLE
+
+    qos_profile.history = QoSHistoryPolicy.RMW_QOS_POLICY_KEEP_ALL_HISTORY if args.keep_all \
+        else QoSHistoryPolicy.RMW_QOS_POLICY_KEEP_LAST_HISTORY
+
+    qos_profile.depth = args.depth
+
+    qos_profile.durability = \
+        QoSDurabilityPolicy.RMW_QOS_POLICY_TRANSIENT_LOCAL_DURABILITY if args.transient_local \
+        else QoSDurabilityPolicy.RMW_QOS_POLICY_VOLATILE_DURABILITY
```
mikaelarguedas: if these end up being 3-line oneliners, maybe using the classic block syntax would be more readable

topic_monitor/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+This demo provides a “topic monitor” that can be used to visualize the statistics of ROS 2 topics that are publishing sequential data.
+
+### How to start publishers
+
+To start one such publisher, assuming you have already installed ROS 2 and sourced your setup file, run:
+```
+topic_monitor_data_publisher critical
+```
+This will start a ROS 2 publisher on the topic `critical_data` that will use the default QoS profile, which includes “reliable” reliability settings. A message of type `std_msgs/Int64` will be sent periodically, with incrementing values that act as sequence numbers. When the topic monitor receives one of these messages, it will maintain an internal counter that is used to keep track of what value should be being received from the publisher. This enables the “reception rate” of the topic to be calculated based on how many of the expected values have been received from the publisher. For a publisher that is communicating with the topic monitor with “reliable” reliability settings, the reception rate is expected to be 100%, even if the topic monitor is running on a different machine, unless that machine is out-of-range of the topic monitor.
```
dirk-thomas: Each sentence should start on a new line.

topic_monitor/launch_files/launch_depth_demo.py
```diff
@@ -0,0 +1,44 @@
...
+        cmd=[executable, 'large', '--depth', '50', '--payload-size', '1000000'],
+    )
+    launch_descriptor.add_process(
+        cmd=[executable, 'large', '--depth', '1', '--payload-size', '1000000'],
+    )
+    launcher.add_launch_descriptor(launch_descriptor)
+
+    rc = launcher.launch()
+    if rc != 0:
+        print('Something went wrong. Return code: ' + str(rc))
```
dirk-thomas: Error messages should go to `stderr`.

Same below.

topic_monitor/launch_files/launch_depth_demo.py
```diff
@@ -0,0 +1,44 @@
...
+        exit()
```
dirk-thomas: This should exit with non-zero return code.

Same below.

topic_monitor/scripts/data_publisher.py
```diff
@@ -0,0 +1,136 @@
...
+        help='Script will exit after publishing this many messages')
+
+    args = parser.parse_args()
+
+    rclpy.init()
+
+    qos_profile = QoSProfile()
+
+    qos_profile.reliability = QoSReliabilityPolicy.RMW_QOS_POLICY_BEST_EFFORT if args.best_effort \
+        else QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABLE
```
dirk-thomas: The `\` is not necessary. Maybe wrap before the `if` for better readability?

Same below.

topic_monitor/scripts/data_publisher.py
```diff
@@ -0,0 +1,136 @@
...
+        msg.frame_id = '%i_%s' % (val, data)
+        data_pub.publish(msg)
+        print('Publishing: "{0}"'.format(val))
+        sys.stdout.flush()  # this is to get the output to show immediately when using Launch
+
+    while rclpy.ok():
+        if args.end_after is not None and cycle_count >= args.end_after:
+            publish_msg(-1)
+            sleep(0.1)  # allow reliable messages to get re-sent if needed
+            exit(0)
```
dirk-thomas: Instead of calling `exit` this could simply `return`. That would make it possible to call this function.

Same below.

topic_monitor/scripts/topic_monitor.py
```diff
@@ -0,0 +1,425 @@
...
+    def check_status(self, current_time=time.time()):
+        if self.status != 'Offline':
+            elapsed_time = current_time - self.time_of_last_data
+            if elapsed_time > self.stale_time * 1.1:
+                self.status_changed = self.status != 'Stale'
+                self.status = 'Stale'
+
+        status_changed = self.status_changed
+        self.status_changed = False
+        return status_changed
```
dhood: The status could have been changed for other reasons: hopefully https://github.com/ros2/demos/pull/90/commits/7c0cb46099b4d7a9a8177c992818449bfaf4d6e0 clarifies this

topic_monitor/scripts/data_publisher.py
```diff
@@ -0,0 +1,125 @@
...
+            return
+
+        publish_msg(cycle_count)
+        cycle_count += 1
+
+        try:
+            sleep(args.period)
+        except KeyboardInterrupt:
+            publish_msg(-1)
+            sleep(0.1)
```
mikaelarguedas: destroying the node and calling shutdown here would avoid ghost DDS entities (like you did in https://github.com/ros2/demos/pull/90/files#diff-00d7c7eb56daa7f13e18b965d5a3572cR296)

wjwwood:
@dhood I'm starting a Linux CI job for this and if it passes I'm going to merge it. I don't want to risk leaving it out to wait for some comments. You can add them later if needed. Here's the CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2096/)

topic_monitor/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    name='topic_monitor',
+    version='0.0.0',
+    packages=[],
+    py_modules=[
+        'launch_files.launch_depth_demo',
+        'launch_files.launch_durability_demo',
+        'launch_files.launch_fragmentation_demo',
+        'launch_files.launch_reliability_demo',
+        'scripts.data_publisher',
+        'scripts.topic_monitor',
```
dirk-thomas: This is not the correct way to install these scripts.

codebot: What is the correct way?

dirk-thomas: The "best" way would be a `console_script` `entry_point` since it generate native executables on Windows. That is what all our Python packages do. A fall back would be to just install the `scripts` (https://docs.python.org/2/distutils/setupscript.html#installing-scripts).

dhood: @dirk-thomas they are installed as `console_script` `entry_point`s at https://github.com/ros2/demos/pull/90/files/cc986ede15e919cc03d8091fe1233d235470ee36#diff-02cee11aa71cd20c15b15b26cc0dfa9bR29 . it is also necessary to install the modules at this line (which is what the other python packages do)
I have tested that they launch fine on all platforms

dirk-thomas: The console scripts look good. But a package should only install code under its own namespace. This package does currently install `scripts` which could collide arbitrarily with any other package.

dhood: that's a good point. I didn't notice because I always symlink install. they get installed under the package's namespace now: https://github.com/ros2/demos/pull/90/commits/1147a7622b3449172bdc3c561049d1a27cf6012b

dhood:
Linux [Build Status](http://ci.ros2.org/job/ci_linux/2100/)
OS X [Build Status](http://ci.ros2.org/job/ci_osx/1622/)
Windows [Build Status](http://ci.ros2.org/job/ci_windows/2114/)

@dirk-thomas I believe your review comment has been addressed but I won't merge this in case that's not true. If you agree, please merge this (I am going offline).

dirk-thomas:
Is it intentional to create "namespace packages" rather then "regular packages" (see [PEP420](https://www.python.org/dev/peps/pep-0420/)) or is that an oversight and the missing `__init__.py` files should be added?

topic_monitor/README.md
```diff
@@ -0,0 +1,86 @@
...
+### Comparing reliability QoS settings
+You will need two machines with ROS 2: one mobile and one stationary.
+
+1. Run the `topic_monitor_launch_reliability_demo` executable on the stationary machine.
+This will start two nodes: one publishing in “reliable” mode, and one in “best effort”.
+1. Start the monitor on a mobile machine such as a laptop.
+Use `topic_monitor --display --allowed-latency 5` to account for any latency that may occur re-sending the reliable messages.
+1. Take the mobile machine out of range of the monitor, and observe how the reception rates differ for the different topics.
+
+You should see that the "reliable" topic has a reception rate that is almost always either 0 or 100%, while the "best effort" topic has a reception rate that fluctuates based on the strength of the connection.
```
dirk-thomas: I am a bit puzzled why this is the case. I would have expected reliable data to arrive (even if with a large latency because of repeated retries) until the connection is completely gone.

dirk-thomas: Could this be related to how this is testing? Is it maybe overloading the bandwidth in that case?

dirk-thomas:
See #98 and #99 for follow up fixes.

PR #93, opened 2016-12-16
force buffer flush and launch file

Karsten1987:


lifecycle/src/lifecycle_listener.cpp
```diff
@@ -68,6 +68,12 @@ class LifecycleListener : public rclcpp::node::Node
 
 int main(int argc, char ** argv)
 {
+  // force flush of the stdout buffer
+  // this ensures a correct sync of all
+  // prints even when executed simultaneously
```
dirk-thomas: Wrapping between "all" and "prints" makes it harder to read the full sentence. Since the line length isn't a problem I would avoid wrapping there and e.g. after "prints" instead.

PR #95, opened 2016-12-17
add server and client

Karsten1987:
Adding server and clients to the composition demo.

I had to do some non-obvious things in the `on_timer` callback for the client. The reason is because we can't trigger AND wait for a service client within a walltimer callback. This effectively results in a spin-within-a-spin which is currently not provided.
The workaround for this is to send the async client request and give it a callback to execute once this request returns true.

@dirk-thomas What is missing here is the test for the two added elements.

### DIFF
--- a/composition/src/client_component.cpp
+++ b/composition/src/client_component.cpp
@@ -47,7 +47,7 @@ Client::Client()
 : Node("Client")
 {
   client_ = create_client<example_interfaces::srv::AddTwoInts>("add_two_ints");
-  timer_ = create_wall_timer(1_s, std::bind(&Client::on_timer, this));
+  timer_ = create_wall_timer(1s, std::bind(&Client::on_timer, this));
 }
 
 bool Client::on_timer()
--- a/composition/src/client_component.cpp
+++ b/composition/src/client_component.cpp
@@ -65,18 +65,23 @@ bool Client::on_timer()
     fprintf(stderr, "service not available, waiting again...\n");
   }
 
-  // We currently cannot call spin in a spin.
-  // That results in the fact that we cannot trigger and wait for
-  // a service call within the timer callback.
-  // The workaround for this is to give the async request another
-  // callback which gets executed once the future is ready.
-  // decltype(client_)::element_type::SharedFuture
-  // = rclcpp::client::Client<example_interfaces::srv::AddTwoInts>::SharedFuture
-  auto return_callback = [](decltype(client_)::element_type::SharedFuture future) {
+  // In order to wait for a response to arrive, we need to spin().
+  // However, this function is already being called from within another spin().
+  // Unfortunately, the current version of spin() is not recursive and so we
+  // cannot call spin() from within another spin().
+  // Therefore, we cannot wait for a response to the request we just made here
+  // within this callback, because it was executed by some other spin function.
+  // The workaround for this is to give the async_send_request() method another
+  // argument which is a callback that gets executed once the future is ready.
+  // We then return from this callback so that the existing spin() function can
+  // continue and our callback will get called once the response is received.
+  using ServiceResponseFuture =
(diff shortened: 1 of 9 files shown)


Karsten1987:
[Build Status](http://ci.ros2.org/job/ci_linux)

composition/src/client_component.cpp
```diff
@@ -0,0 +1,93 @@
...
+    status = future.wait_for((time_left < wait_period) ? time_left : wait_period);
+  } while (rclcpp::ok() && status != std::future_status::ready);
+  return status;
+}
+
+Client::Client()
+: Node("Client")
+{
+  client_ = create_client<example_interfaces::srv::AddTwoInts>("add_two_ints");
+  timer_ = create_wall_timer(1_s, std::bind(&Client::on_timer, this));
```
wjwwood: Actually, this needs to be `1s` (which is how you make use of `using namespace std::chrono_literals`).

wjwwood:
I addressed my last comment in 199452b, but I'm going to wait to redo CI until after I can merge the time literals related pr's, given the nature of 199452b.

Karsten1987:
CI passes all tests.
The compiler warning comes from the tutorial package.

Karsten1987:
@dirk-thomas Would you mind having a look at it?

composition/src/client_component.cpp
```diff
@@ -0,0 +1,93 @@
...
+  // Therefore, we cannot wait for a response to the request we just made here
+  // within this callback, because it was executed by some other spin function.
+  // The workaround for this is to give the async_send_request() method another
+  // argument which is a callback that gets executed once the future is ready.
+  // We then return from this callback so that the existing spin() function can
+  // continue and our callback will get called once the response is received.
+  using ServiceResponseFuture =
+      rclcpp::client::Client<example_interfaces::srv::AddTwoInts>::SharedFuture;
+  auto response_received_callback = [](ServiceResponseFuture future) {
+      printf("Got result: [%s]\n",
```
dirk-thomas: Why use `%s` instead of `%zu` (or whatever platform specific macro matches the type in the message)?

Same below.

wjwwood: I think this is one of those cases where you have to use the `PRI...` macros instead of a `%XX` because the type (`uint64_t` in this case I think) is actually different depending on the platform.

wjwwood:
I think he's got passing CI on Linux and passing on the others (minus a style failure which he fixed) already:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2093/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1619/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2110/)

(Windows actually also had a flaky test failure).

Either way he has Linux passing, so I'm going to merge this and let the nightlies chew on it. @Karsten1987 if you decide to address the printf stuff, please open a follow up pr.

dirk-thomas:
This might be responsible for failing the composition test across all platforms in the last nightlies: e.g. http://ci.ros2.org/view/nightly/job/nightly_linux_debug/302/testReport/

Karsten1987:
I have the feeling that the regex test is flaky. And adding a second line for the services doesn't help in that case. I believe this is a race condition between the talker/listener/server/client and depending on the order they start, the stdout output is differently. 
This test was never flaky before my additions?

dirk-thomas:
I didn't look flaky to me. It failed in all nine nightly jobs. And six of them even try multiple times and it failed everytime.

The output suggests that the output handler already aborts the test after the listener has received the message. Have you checked how the two lines in the regex are being checked? What logic applies to them?

dirk-thomas:
I haven't seen the test fail since it has been added.

PR #97, opened 2016-12-19
Add comments to composition demo

dhood:


composition/src/listener_component.cpp
```diff
@@ -23,16 +23,31 @@
...
+ * Components get built into shared libraries and as such do not write their own main functions.
+ * The process using the component's shared library will instantiate the class as a ROS node.
+ */
 Listener::Listener()
 : Node("listener")
 {
+  /**
+    * Create a callback function for when messages are received.
+    * Variations of this function also exist using, for example, UniquePtr for zero-copy transport.
+    */
```
dirk-thomas: Should these be line comments rather then block comments. Doxygen (`/**`) would also not extract those I assume?

dirk-thomas: Same for the other files.

wjwwood:
I updated this to use line comments in order to fix the inconsistent indentation and to avoid confusion with doxygen and to save vertical lines. I don't think we were planning on running doxygen on this anyways.

dirk-thomas:
Still one file with block comments.

wjwwood:
It's fixed.

dhood:
Block comments are used in ROS 1 tutorials so I was following that
Thank you for iterating on this @wjwwood!

wjwwood:
> Block comments are used in ROS 1 tutorials so I was following that

Ah, that makes sense. I was just updating it to follow our developer guide:

> Use /// and /** */ comments for documentation purposes and // style comments for notes and general comments
>
> -- https://github.com/ros2/ros2/wiki/Developer-Guide#c-1

Both paths were probably fine, but I switched it when touching things to fix the indentation issues.

PR #101, opened 2016-12-19
add missing shebang line and fix permissions

### LINKED ISSUE #100: [lifecycle] installed python files are not executable
They are in `bin`, but not executable and so `which` doesn't find them:

```
% which lifecycle_listener
/tmp/testing_beta1/ros2-osx/bin/lifecycle_listener
% which lifecycle_demo_launch.py
lifecycle_demo_launch.py not found
```

However, this is how the user is recommended to execute them:

> ```
> python3 `which lifecycle_service_client_py.py`
> ```
>
> -- https://github.com/ros2/ros2/wiki/Managed-Nodes

Not sure if this works on Linux as-is, but it doesn't work with `zsh` on OS X (my machine).

dirk-thomas:
Fixes #100.

### DIFF
--- a/lifecycle/CMakeLists.txt
+++ b/lifecycle/CMakeLists.txt
@@ -59,7 +59,7 @@ endif()
 
 # install python demo
 install(
-  FILES
+  PROGRAMS
   src/lifecycle_service_client_py.py
   launch/lifecycle_demo_launch.py
   DESTINATION bin)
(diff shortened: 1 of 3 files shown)


Karsten1987:
This works for me. On my machine (linux) I can either run
``` python3 `which lifecycle_demo_launch.py` ```
or directly
``` lifecycle_demo_launch.py```

wjwwood:
Fixed on OS X too.

PR #103, opened 2016-12-21
Print out errors reported the result of set parameter

rohbotics:


demo_nodes_cpp/src/parameters/ros2param.cpp
```diff
@@ -139,6 +139,12 @@ int main(int argc, char ** argv)
       fprintf(stderr, "Failed to set parameter\n");
       return 1;
     }
+    auto results = set_parameters_result.get();
+    for (auto result : results) {
+      if (result.successful != true) {
```
dirk-thomas: Instead of `!= true` the expression should be prefixed with `!`.

rohbotics: The rest of the file is using `!=`, should I change those as well, just this, or no change to keep it consistent?

dirk-thomas: i don't see any other line using `!= true`. Can you please identify the lines you are referring to with links.

rohbotics: I was looking at https://github.com/rohbotics/demos/blob/948ca1ef6b228e7aea94eff2b6b761ea0b2c2bfc/demo_nodes_cpp/src/parameters/ros2param.cpp#L138 

And https://github.com/rohbotics/demos/blob/948ca1ef6b228e7aea94eff2b6b761ea0b2c2bfc/demo_nodes_cpp/src/parameters/ros2param.cpp#L113

But those aren't comparing to `true`, to I changed this line.

demo_nodes_cpp/src/parameters/ros2param.cpp
```diff
@@ -139,6 +139,12 @@ int main(int argc, char ** argv)
       fprintf(stderr, "Failed to set parameter\n");
       return 1;
     }
+    auto results = set_parameters_result.get();
+    for (auto result : results) {
+      if (!result.successful = true) {
```
dirk-thomas: This logic is wrong. Needs to be just `!result.successful`.

demo_nodes_cpp/src/parameters/ros2param.cpp
```diff
@@ -139,6 +139,12 @@ int main(int argc, char ** argv)
       fprintf(stderr, "Failed to set parameter\n");
       return 1;
     }
+    auto results = set_parameters_result.get();
+    for (auto result : results) {
+      if (!result.successful) {
+        fprintf(stderr, "Error setting parameter: %s\n", result.reason.c_str());
```
dirk-thomas: It would be useful if the error message would mention which parameters failed to set.

rohbotics: The command line interface only supports one parameter at a time, but there is no API for setting one parameter.

dirk-thomas: Then this code shouldn't need to iterate the results since there must always be exactly one.

wjwwood:
I would say we should run CI, but I think they're broken right now due to a recent Fast-RTPS change.

PR #105, opened 2017-01-05
use rmw implementation

dirk-thomas:
Connect to ros2/rcl#106

### DIFF
--- a/composition/CMakeLists.txt
+++ b/composition/CMakeLists.txt
@@ -11,16 +11,9 @@ find_package(ament_index_cpp REQUIRED)
 find_package(class_loader REQUIRED)
 find_package(example_interfaces REQUIRED)
 find_package(rclcpp REQUIRED)
-find_package(rmw_implementation_cmake REQUIRED)
 find_package(rosidl_cmake REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
 find_package(std_msgs REQUIRED)
-
-# get the rmw implementations ahead of time
-find_package(rmw_implementation_cmake REQUIRED)
-get_available_rmw_implementations(rmw_implementations)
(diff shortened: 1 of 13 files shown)


dirk-thomas:
Updating the docs etc. is listed at the very end of ros2/ros2#302 as pending (after everything is updated and merged). I think the wiki should track *master* since otherwise there is no documentation for *master*  (the state for beta 1 can always be retrieved from a hash of the wiki pages) and before the next release we would need to update multiple things at once.

wjwwood:
While it's possible to get the beta1 version of the wiki from commit hashes, it isn't very easy to do that. So I think this underscores the need to eventually move towards something that can be version controlled. Whether that is something like wiki.ros.org seesaws, or everything statically generated to different folders (sphinx and/or doxygen) or something like gazebosim.org's versioned tutorials (http://gazebosim.org/tutorials?tut=install_ubuntu&cat=install).

dirk-thomas:
I absolutely agree. But since that is likely unrealistic for beta 2 we need to decide what to do until then.

wjwwood:
I'm looking to see if there is a way to host some version of the wiki statically by cloning it and rendering it. This project looks promising: https://github.com/joeyespo/grip It is designed to render README's, but it uses the GitHub API to do the rendering which means it gets cross links and stuff.

Another option might be to duplicate all of the wiki pages into a `beta1` folder locally and see if that lets us easily take a snapshot (though the inter-wiki links might be broken then).

Just note the current commit hash of the wiki in the ticket you referenced before you make these changes, if you don't mind. That way we can continue to look at some way to provide both to users.

dirk-thomas:
I simple approach for now would be to add a note at the beginning of a package with a link to the beta 1 version of that page. Easily done, no further effort. :wink:

PR #116, opened 2017-02-28
Add support for RGB8

clalancette:
The showimage viewer is a useful program to have for viewing data other than that from cam2image.  Here we expand it so that it supports rgb8 as well as bgr8.

### DIFF
--- a/image_tools/src/showimage.cpp
+++ b/image_tools/src/showimage.cpp
@@ -34,15 +34,20 @@ encoding2mat_type(const std::string & encoding)
 {
   if (encoding == "mono8") {
     return CV_8UC1;
-  } else if (encoding == "bgr8") {
+  }
+  else if (encoding == "bgr8") {
     return CV_8UC3;
-  } else if (encoding == "mono16") {
+  }
+  else if (encoding == "mono16") {
     return CV_16SC1;
-  } else if (encoding == "rgba8") {
+  }
+  else if (encoding == "rgba8") {
     return CV_8UC4;
-  } else if (encoding == "32FC1") {
+  }
+  else if (encoding == "32FC1") {
     return CV_32FC1;
-  } else {
+  }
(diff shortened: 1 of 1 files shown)


image_tools/src/showimage.cpp
```diff
@@ -34,15 +35,23 @@ encoding2mat_type(const std::string & encoding)
 {
   if (encoding == "mono8") {
     return CV_8UC1;
-  } else if (encoding == "bgr8") {
+  }
+  else if (encoding == "bgr8") {
```
mikaelarguedas: this changes will make uncrustify unhappy. Is it possible to revert them?

mikaelarguedas: Tip: uncrustify / ament_uncrustify support the `--reformat` flag. It will lint the code for you ;)

image_tools/src/showimage.cpp
```diff
@@ -81,9 +99,8 @@ int main(int argc, char * argv[])
   bool show_camera = true;
 
   // Configure demo parameters with command line options.
-  if (!parse_command_options(argc, argv, &depth, &reliability_policy, &history_policy,
-    &show_camera))
-  {
+  if (!parse_command_options(argc, argv, &depth, &reliability_policy,
+                             &history_policy, &show_camera)) {
```
mikaelarguedas: we usually wrap lines right after the opening parenthesis rather than between parameters (c.f. [Developer guide](https://github.com/ros2/ros2/wiki/Developer-Guide#c-1))

clalancette:
Ah, got it.  I just pushed a new patch here that makes uncrustify happy.  Let me know if this is better.

mikaelarguedas:
looks good to me :+1:

clalancette:
Thanks, I'll merge this in.

PR #117, opened 2017-03-07
Allow showimage and cam2image to subscribe/publish to alternate topics.

clalancette:
This allows them to be a little more generic and useful for debugging.  In particular, this is helpful for debugging the Astra camera, as we can have showimage subscribe to either the RGB or the depth image without a recompile.

### DIFF
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -114,7 +114,7 @@ int main(int argc, char * argv[])
   // parameter.
   custom_camera_qos_profile.history = history_policy;
 
-  std::cout << "Publishing data on topic " << topic << std::endl;
+  printf("Publishing data on topic %s\n", topic.c_str());
   // Create the image publisher with our custom QoS profile.
   auto pub = node->create_publisher<sensor_msgs::msg::Image>(
     topic, custom_camera_qos_profile);
@@ -195,7 +195,7 @@ int main(int argc, char * argv[])
         cv::waitKey(1);
       }
       // Publish the image message and increment the frame_id.
-      std::cout << "Publishing image #" << i << std::endl;
+      printf("Publishing image #%zd\n", i);
       pub->publish(msg);
       ++i;
     }
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -114,7 +114,7 @@ int main(int argc, char * argv[])
   // parameter.
   custom_camera_qos_profile.history = history_policy;
 
-  printf("Publishing data on topic: %s\n", topic.c_str());
+  printf("Publishing data on topic '%s'\n", topic.c_str());
   // Create the image publisher with our custom QoS profile.
   auto pub = node->create_publisher<sensor_msgs::msg::Image>(
     topic, custom_camera_qos_profile);
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -114,7 +114,7 @@ int main(int argc, char * argv[])
   // parameter.
   custom_camera_qos_profile.history = history_policy;
 
-  printf("Publishing data on topic %s\n", topic.c_str());
+  printf("Publishing data on topic: %s\n", topic.c_str());
   // Create the image publisher with our custom QoS profile.
   auto pub = node->create_publisher<sensor_msgs::msg::Image>(
     topic, custom_camera_qos_profile);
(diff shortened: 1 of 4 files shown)


image_tools/src/cam2image.cpp
```diff
@@ -193,7 +194,7 @@ int main(int argc, char * argv[])
         cv::waitKey(1);
       }
       // Publish the image message and increment the frame_id.
-      std::cout << "Publishing image #" << i << std::endl;
+      std::cout << "Publishing image #" << i << " to topic " << topic << std::endl;
```
dhood: If you could instead output the topic once on startup, the tests for these executables shouldn't have to be updated https://github.com/ros2/demos/blob/ec3affc96896f338e125be8ba553de92851956f2/image_tools/test/cam2image.regex#L1
(the tests are disabled but will be re-enabled in https://github.com/ros2/demos/pull/83)

clalancette: Oh, sure, no problem.  I'll do that.

image_tools/src/cam2image.cpp
```diff
@@ -113,9 +114,10 @@ int main(int argc, char * argv[])
   // parameter.
   custom_camera_qos_profile.history = history_policy;
 
+  std::cout << "Publishing data on topic " << topic << std::endl;
```
dirk-thomas: Please use `printf` instead (see https://github.com/ros2/ros2/wiki/Developer-Guide#c-specific).

image_tools/src/options.cpp
```diff
@@ -76,6 +77,9 @@ bool parse_command_options(
     if (burger_mode != nullptr) {
       ss << " -b: produce images of burgers rather than connecting to a camera" << std::endl;
     }
+    if (topic != nullptr) {
+      ss << " -t TOPIC: use topic TOPIC instead of the default \"image\"" << std::endl;
```
dirk-thomas: It would be good not to repeat the default value.

image_tools/src/cam2image.cpp
clalancette: I'm not opposed to switching it to printf() (I generally prefer that anyway), but I'm surprised at the rationale.  printf() doesn't guarantee atomicity when writing data out, so it can also be interleaved when multiple threads are in play.  Is the C++ iostream considerably different here?

image_tools/src/options.cpp
clalancette: I agree duplicating it here isn't ideal, and with the added print at publisher creation time we don't need it anymore.  To be honest, this entire option parsing function could be done better; I think it would be nicer to use/write some sort of library that is more similar to Python's argparse.  In any case, I'll change this.

image_tools/src/cam2image.cpp
dirk-thomas: I think it depends on the implementation of the C library. For POSIX the following is stated in http://pubs.opengroup.org/onlinepubs/9699919799/functions/flockfile.html :

> All functions that reference (FILE *) objects, except those with names ending in _unlocked, shall behave as if they use flockfile() and funlockfile() internally to obtain ownership of these (FILE *) objects.

I don't think I have ever seen a single `printf` statement being interleaved with another `printf` statement from a different thread. For `cout << "foo" << "bar" << endl` this will certainly be written in three separate steps and interleaving is quite common for those.

clalancette: Huh, OK, I didn't know about that piece of language from POSIX.  I've switched it to use printf now.

dhood:
when I was testing this locally I thought `Subscribing to topic image` sounded confusing so I added https://github.com/ros2/demos/pull/117/commits/a264625a3846289fa16efbaa37714fbd3a1f3a45. otherwise works well locally so LGTM

mikaelarguedas:
Nice! We usually quote with simple quotes (`'%s'`) for clarity, is it possible to match this pattern here ?

dhood:
> Nice! We usually quote with simple quotes ('%s') for clarity, is it possible to match this pattern here ?


https://github.com/ros2/demos/pull/117/commits/feaa386df476de3b4ec630b385a9ac8ad5b1100d

image_tools/src/showimage.cpp
```diff
@@ -130,7 +130,7 @@ int main(int argc, char * argv[])
       show_image(msg, show_camera);
     };
 
-  std::cout << "Subscribing to topic: " << topic << std::endl;
+  printf("Subscribing to topic '%s'\n", topic.c_str());
```
mikaelarguedas: thanks :+1:

clalancette:
Thanks everyone, I've merged this now.

PR #119, opened 2017-03-24
use ament_cmake_export_interfaces

dirk-thomas:
Uses the interface exported by `ament_index_cpp`. See ament/ament_index#22.

Connect to ament/ament_index#22

### DIFF
--- a/composition/CMakeLists.txt
+++ b/composition/CMakeLists.txt
@@ -104,11 +104,11 @@ ament_target_dependencies(dlopen_composition
 
 add_executable(api_composition
   src/api_composition.cpp)
+target_link_libraries(api_composition ament_index_cpp::ament_index_cpp)
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Waiting for review.

composition/CMakeLists.txt
```diff
@@ -104,11 +104,11 @@ ament_target_dependencies(dlopen_composition
 
 add_executable(api_composition
   src/api_composition.cpp)
+target_link_libraries(api_composition ament_index_cpp::ament_index_cpp)
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
   target_link_libraries(api_composition "stdc++fs")
 endif()
 ament_target_dependencies(api_composition
-  "ament_index_cpp"
```
wjwwood: Is this completely replaced by the target interface? What about compiler definitions and linker flags?

dirk-thomas: Yes, everything necessary is encapsulated in the imported target.

PR #129, opened 2017-05-19
Fixed incorrect image size in watermark and image view nodes

Kukanani:
cv::Mat constructor takes (# of rows, # of cols), which is (height, width) instead of the previously used (width, height).

This fix causes images and watermarks to look right when running demos like image_pipeline_all_in_one.


mikaelarguedas:
Tested locally and it does fix the display. Running CI after rebasing just in case:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2627/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/189/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2738/)

wjwwood:
Or I should say I fixed a similar issue in the other demo.

PR #130, opened 2017-05-19
Added ability to disable pid/ptr on images

Kukanani:
For image_view_node and camera_node, added flag that turns off the "watermark" with the pid and image
data pointer. The default behavior remains unchanged.


intra_process_demo/src/image_pipeline/camera_node.hpp
```diff
@@ -90,6 +93,10 @@ class CameraNode : public rclcpp::Node
   std::thread thread_;
   std::atomic<bool> canceled_;
 
+  /// whether or not to add a watermark to the image showing process id and
+  /// pointer location
+  std::atomic<bool> watermark_;
```
wjwwood: Are you swapping this during runtime? If not, then it probably doesn't need to be atomic, but if you are, then it is fine as-is.

PR #131, opened 2017-05-19
Move header files from intra_process_demo into an include directory

Kukanani:
This makes it cleaner for other packages to depend on some nodes contained in image_pipeline, namely camera_node and image_view_node. 

Moved image_pipeline header files into include directory and updated CMakeLists accordingly


intra_process_demo/CMakeLists.txt
```diff
@@ -84,6 +85,12 @@ install(TARGETS
   image_view_node
   DESTINATION bin)
 
+install(
+  DIRECTORY include/
+  DESTINATION include
+)
+
+
```
wjwwood: nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

PR #133, opened 2017-06-02
return code 0 when sigint is received while waiting for response

### LINKED ISSUE #132: api composition demo test is flakey
See this failure for example:

http://ci.ros2.org/job/nightly_linux-aarch64_repeated/57/testReport/(root)/projectroot/test_composition__rmw_fastrtps_cpp/
```
(test_api_composition) pid 32385: ['/home/rosbuild/ci_scripts/ws/build/composition/api_composition'] (all > console, InMemoryHandler: test_api_composition)
(load_talker_component) pid 32386: ['/home/rosbuild/ci_scripts/ws/build/composition/api_composition_cli', 'composition', 'composition::Talker'] (stderr > stdout, all > console)
(load_listener_component) pid 32387: ['/home/rosbuild/ci_scripts/ws/build/composition/api_composition_cli', 'composition', 'composition::Listener'] (stderr > stdout, all > console)
[load_listener_component] Sending request...
[load_listener_component] Waiting for response...
[load_listener_component] Result of load_node: success = true
(load_listener_component) rc 0
[test_api_composition] Load library /home/rosbuild/ci_scripts/ws/build/composition/liblistener_component.so
[test_api_composition] Instantiate class composition::Listener
[test_api_composition] I heard: [Hello World: 58]
[test_api_composition] signal_handler(2)
(test_api_composition) rc 0
() tear down
(load_talker_component) signal SIGINT
[load_talker_component] api_composition_cli was interrupted. Exiting.
[load_talker_component] Sending request...
[load_talker_component] Waiting for response...
[load_talker_component] signal_handler(2)
(load_talker_component) rc 1
```

This is what is happening, in my opinion:

- `test_api_composition` starts and launch is looking for `I heard: [Hello World:` before assuming success of the test and sending sigint
- `load_listener_component` loads the listener into the "container" `test_api_composition`, is successful and exits clean
- `load_talker_component` also loads the talker into the container, but does not exit right away
- the talker and listener talk with each other, launch notices and sends everyone sigint (success!)
- but since `load_talker_component` isn't "done" yet (has not received the service response yet) the sigint it receives is interpreted as an error

Possible solution:

- change launch to wait on sending sigint until it sees `I heard: [Hello World:` and `Result of load_node: success = true` two times

I don't know how to do that off hand, and I don't have time to follow up on it right now, but maybe sometime this week or next while I'm still build cop.

If anyone has any idea about how to make launch do what I described above, let me know.

dirk-thomas:
This should fix #132. There are other cases where this test is flaky though.

### DIFF
--- a/composition/src/api_composition_cli.cpp
+++ b/composition/src/api_composition_cli.cpp
@@ -51,6 +51,9 @@ int main(int argc, char * argv[])
     rclcpp::executor::FutureReturnCode::SUCCESS)
   {
     fprintf(stderr, "api_composition_cli was interrupted. Exiting.\n");
+    if (!rclcpp::ok()) {
+      return 0;
+    }
     return 1;
   }
   printf("Result of load_node: success = %s\n", result.get()->success ? "true" : "false");


composition/src/api_composition_cli.cpp
```diff
@@ -51,6 +51,9 @@ int main(int argc, char * argv[])
     rclcpp::executor::FutureReturnCode::SUCCESS)
   {
     fprintf(stderr, "api_composition_cli was interrupted. Exiting.\n");
+    if (!rclcpp::ok()) {
+      return 0;
+    }
```
wjwwood: Isn't this case always true?

wjwwood:
> There are other cases where this test is flaky though.

How else? Are there issues which describe how it is flakey?

composition/src/api_composition_cli.cpp
dirk-thomas: In the case where the future contains not success (and this script hasn't been asked to terminate using a SIGINT) I think this script should return a non-zero code.

dirk-thomas:
> How else? Are there issues which describe how it is flakey?

Other jobs where this test fails don't show the output you mentioned in the first comment. Some don't generate any output so I don't think they all have the same reason.

dirk-thomas:
I will go ahead and merge this. If you still think something should be changed please feel free to comment and I will try to address it.

PR #141, opened 2017-06-26
add parameter node to test ros2param

dirk-thomas:


demo_nodes_cpp/src/parameters/parameter_node.cpp
```diff
@@ -0,0 +1,57 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
```
wjwwood: I know this is based on another demo, but I think the copyright should be updated to 2017 since this is a copy not a move.

dirk-thomas: Why would the year of the copyright change when copying the file?

wjwwood: Because it's a new file, and not a relocation of an existing file? Whether or not it is written from scratch or based on something else and modified shouldn't matter imo.

You can do whatever you want, but it seems like it should be updated to me.

dirk-thomas: My rational is the following: if a file was written by Alice in 2016 and in 2017 copied by Bob to a new location the copyright is clearly still lying with Alice in 2016. Bob could only add his copyright and the current year if he did some non-trivial changes.

PR #149, opened 2017-06-28
Update topic monitor for get_topic_names_and_types changes

dhood:
I haven't moved the demo bringup files into actual `launch` files because otherwise you need invocations the equivalent of:
```
launch `ros2 pkg prefix dummy_robot_bringup`/share/dummy_robot_bringup/launch/dummy_robot_bringup.py
```
to get them to run, and for now `ros2 run topic_monitor launch_*_demo` is cleaner for the readme. I understand [from comments like this](https://github.com/ros2/turtlebot2_demo/pull/44#issuecomment-310739512) that the launch api is not really ready to be used yet anyway, so at least with `ros2 run` there are not paths to deal with.

### DIFF
--- a/topic_monitor/topic_monitor/scripts/topic_monitor.py
+++ b/topic_monitor/topic_monitor/scripts/topic_monitor.py
@@ -109,7 +109,7 @@ class TopicMonitor:
     """Monitor of a set of topics that match a specified topic name pattern."""
 
     def __init__(self, window_size):
-        self.data_topic_pattern = re.compile("((?P<data_name>\w*)_data_?(?P<reliability>\w*))")
+        self.data_topic_pattern = re.compile("(/(?P<data_name>\w*)_data_?(?P<reliability>\w*))")
         self.monitored_topics = {}
         self.monitored_topics_lock = Lock()
         self.publishers = {}
@@ -143,7 +143,7 @@ class TopicMonitor:
 
         # Create a publisher for the reception rate of the topic
         reception_rate_publisher = node.create_publisher(
-            Float32, self.reception_rate_topic_name + '/' + topic_name)
+            Float32, self.reception_rate_topic_name + topic_name)
 
         with self.monitored_topics_lock:
             monitored_topic.expected_value_timer = expected_value_timer
--- a/topic_monitor/topic_monitor/scripts/topic_monitor.py
+++ b/topic_monitor/topic_monitor/scripts/topic_monitor.py
@@ -143,8 +143,15 @@ class TopicMonitor:
         allowed_latency_timer.cancel()
 
         # Create a publisher for the reception rate of the topic
+        reception_rate_topic_name = self.reception_rate_topic_name + topic_name
+
+        # TODO(dhood): remove this workaround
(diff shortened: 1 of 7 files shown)


topic_monitor/topic_monitor/scripts/topic_monitor.py
```diff
@@ -306,7 +306,10 @@ def run_topic_listening(node, topic_monitor, options):
         # TODO(dhood): use graph events rather than polling
         topic_names_and_types = node.get_topic_names_and_types()
 
-        for topic_name, type_name in topic_names_and_types:
+        for topic_name, type_names in topic_names_and_types:
+            if len(type_names) != 1:
+                continue
```
wjwwood: I recommend putting a message here, otherwise there is no feedback to the user that a topic it is monitoring is being ignored.

In the bridge I kept a list of topic names for which a similar warning was printed and used it to only print once.

dhood: ok, will do

dirk-thomas:
Since you create the `__init__.py` files I would recommend to not install the modules using `py_modules` in the `setup.py` file anymore. Instead install the whole package (like here: https://github.com/ament/ament_index/blob/423d16f2dfb5821f828efc5202b38444e0482817/ament_index_python/setup.py#L7).

dhood:
Thanks @dirk-thomas and @wjwwood for setup.py guidance

I added a workaround for connext in https://github.com/ros2/demos/pull/149/commits/3e14d66a7f8a6dafeed77fa392be79f93b6f2ead that must be needed since the namespace implementation's been merged. Ticketed in https://github.com/ros2/rmw_connext/issues/234

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2844/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/348/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2300/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2981/)

re-running with launch as exec_depend:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2845/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/349/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2301/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2982/)

topic_monitor/topic_monitor/scripts/topic_monitor.py
```diff
@@ -362,8 +385,11 @@ def main():
 
     args = parser.parse_args()
     if args.show_display:
-        global plt
-        import matplotlib.pyplot as plt
+        try:
+            global plt
+            import matplotlib.pyplot as plt
+        except:
```
mikaelarguedas: Nit: a more precise way would be to catch `ImportError` here

dirk-thomas: Done in 6a03fc5f979c4bb93215e580cd218751aedea971

PR #150, opened 2017-06-29
[demo_nodes_py] service/client mimic c++ demo behavior

### LINKED ISSUE #145: add_two_ints in python and c++ should have the same behaviour
the c++ server stays upe and the client sends only one request at a tine
the python server shuts down after 3 requests and the client send them by 3-request batches

mikaelarguedas:
Decided to adress this when debugging services on Windows with @tfoote 

The first commit Fixes #145 
The second commit calls `rclpy.shutdown()` at the end of each executable. (I didnt add explicit calls to node.destroy_node() because that should have been fixed by rclpy https://github.com/ros2/rclpy/pull/94)

### DIFF
--- a/demo_nodes_py/services/add_two_ints_client.py
+++ b/demo_nodes_py/services/add_two_ints_client.py
@@ -23,16 +23,15 @@ def main(args=None):
     node = rclpy.create_node('add_two_ints_client')
 
     cli = node.create_client(AddTwoInts, 'add_two_ints')
-    max_iter = 3
-    i = 0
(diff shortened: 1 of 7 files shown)


mikaelarguedas:
Actually https://github.com/ros2/rclpy/pull/94 doesn't fix it just report the error properly. the third commit add a call to `destroy_node()` before shutdown so that the node handle is still valid when trying to destroy it (otherwise shutdown does shut everything down and thus the nodes handles are not valid when the Python node object goes out of scope and tries to destroy itself)

PR #153, opened 2017-07-05
add missing runtime dependencies

mikaelarguedas:


mikaelarguedas:
ready for review

Karsten1987:
lgtm. I guess there are problems when packaging ?

mikaelarguedas:
Packaging does pass but running fails because the launchfile is trying to spawn these nodes while they don't exist (because not installed if we don't declare these deps)

mikaelarguedas:
I'll take https://github.com/ros2/demos/pull/153#issuecomment-313472084 as an approval and will go ahead and merge this

PR #157, opened 2017-08-05
pep8 + import order

mikaelarguedas:
noticed style when reviewing a PR in this repo

### DIFF
--- a/composition/test/test_composition.py.in
+++ b/composition/test/test_composition.py.in
@@ -48,11 +48,13 @@ def test_dlopen_composition():
     server_library = '@SERVER_LIBRARY@'
     client_library = '@CLIENT_LIBRARY@'
     output_file = '@EXPECTED_OUTPUT_ALL@'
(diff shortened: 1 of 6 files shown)


mikaelarguedas:
well we could easily [extend ament_flake8](https://github.com/ament/ament_lint/compare/flake8_check_template_files) to process template files as well.
We may have to change some of them that are valid templates but that flake8 will not like e.g. https://github.com/ros2/rcl/blob/2811eb1312ce435ef891c8a0308c4e0ef0633bf5/rcl/test/rcl/test_two_executables.py.in#L9

mikaelarguedas:
I'm going to merge this but feel free to continue the discussion here

dirk-thomas:
I don't think that it is viable to lint the templates. Instead the expanded template could be linted (we do the same for the generated code of messages/services).

mikaelarguedas:
Sounds reasonable to me, feel free to open a ticket with a suggested approach because I'm not sure of what would be a good way to register files expanded by CMake for linting

PR #158, opened 2017-08-07
remove dependency on ament_python and perform customizations in setup.py

dirk-thomas:
Connect to ament/ament_python#3.

The same changes are necessary for the `example_rclpy_*` packages. Once we agree this is the right way to move forward I will create a PR for those.

**Update:** ros2/examples#178

### DIFF
--- a/demo_nodes_py/setup.py
+++ b/demo_nodes_py/setup.py
@@ -1,17 +1,17 @@
-from ament_python.data_files import get_data_files
-from ament_python.script_dir import install_scripts_to_libexec
 from setuptools import find_packages
 from setuptools import setup
 
 package_name = 'demo_nodes_py'
-data_files = get_data_files(package_name)
-install_scripts_to_libexec(package_name)
 
 setup(
     name=package_name,
     version='0.0.0',
     packages=find_packages(exclude=['test']),
-    data_files=data_files,
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
+    ],
     install_requires=['setuptools'],
     author='Esteve Fernandez',
     author_email='esteve@osrfoundation.org',
--- a/topic_monitor/setup.py
+++ b/topic_monitor/setup.py
@@ -1,17 +1,17 @@
-from ament_python.data_files import get_data_files
-from ament_python.script_dir import install_scripts_to_libexec
 from setuptools import find_packages
 from setuptools import setup
 
 package_name = 'topic_monitor'
-data_files = get_data_files(package_name)
-install_scripts_to_libexec(package_name)
 
 setup(
-    name='topic_monitor',
+    name=package_name,
     version='0.0.0',
     packages=find_packages(exclude=['test']),
-    data_files=data_files,
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
+    ],
     install_requires=[
         'launch',
         'setuptools',
--- a/demo_nodes_py/package.xml
+++ b/demo_nodes_py/package.xml
@@ -10,8 +10,6 @@
   <maintainer email="mikael@osrfoundation.org">Mikael Arguedas</maintainer>
   <license>Apache License 2.0</license>
 
-  <build_depend>ament_python</build_depend>
-
   <exec_depend>example_interfaces</exec_depend>
   <exec_depend>rclpy</exec_depend>
   <exec_depend>std_msgs</exec_depend>
--- a/demo_nodes_py/setup.cfg
+++ b/demo_nodes_py/setup.cfg
@@ -1,0 +1,4 @@
+[develop]
+script-dir=$base/lib/demo_nodes_py
+[install]
+install-scripts=$base/lib/demo_nodes_py
--- a/topic_monitor/package.xml
+++ b/topic_monitor/package.xml
@@ -7,7 +7,6 @@
   <maintainer email="dhood@osrfoundation.org">D. Hood</maintainer>
   <license>Apache License 2.0</license>
 
-  <build_depend>ament_python</build_depend>
   <build_depend>rclpy</build_depend>
 
   <exec_depend>launch</exec_depend>
--- a/topic_monitor/setup.cfg
+++ b/topic_monitor/setup.cfg
@@ -1,0 +1,4 @@
+[develop]
+script-dir=$base/lib/topic_monitor
+[install]
+install-scripts=$base/lib/topic_monitor


demo_nodes_py/setup.py
```diff
@@ -1,17 +1,27 @@
...
 from setuptools import find_packages
 from setuptools import setup
 
 package_name = 'demo_nodes_py'
-data_files = get_data_files(package_name)
-install_scripts_to_libexec(package_name)
+
+# install scripts into libexec path
+if 'develop' in sys.argv:
+    sys.argv[sys.argv.index('develop') + 1:sys.argv.index('develop') + 1] = \
```
nuclearsandwich: This appears to insert the arguments but I don't see how it replaces a script-dir argument already part of argv the way that the ament_python version of this code did. https://github.com/ament/ament_python/blob/eafd268376cda400f68b1b5b8429986531b856ad/ament_python/script_dir.py#L27-L29

Same for the install command just below and the install-scripts argument which we do specify during debian packaging in order to make sure the non-libexec scripts are placed into `/opt/ros/$ROSDISTRO/bin`

topic_monitor/setup.py
```diff
@@ -1,17 +1,27 @@
...
 from setuptools import find_packages
 from setuptools import setup
 
 package_name = 'topic_monitor'
-data_files = get_data_files(package_name)
-install_scripts_to_libexec(package_name)
+
+# install scripts into libexec path
+if 'develop' in sys.argv:
+    sys.argv[sys.argv.index('develop') + 1:sys.argv.index('develop') + 1] = \
```
nuclearsandwich: Same question as above (link to be edited in pending submission). How does this replace a manually specified script-dir argument?

dirk-thomas: It doesn't. But I don't think it needs to either. ament/ament_tools#156 removes passing those from the build tool. So if no custom location is passed the package installs it's scripts into libexec. If a custom location is passed it is being honored (imo better than being ignored/overridden as it was before). The Debian rules file will need to do the same as the build tool (pass a prefix but no script specific location) and the behavior should not change.

nuclearsandwich: > The Debian rules file will need to do the same as the build tool (pass a prefix but no script specific location) and the behavior should not change.

I'm not sure that this will work. Earlier iterations of the bloom changes passed only the prefix and that was not sufficient to put everything in the right place. I needed to pass both --install-lib and --install-scripts explicitly. Blooming this branch in a throwaway repository, patching out the explicit --install-scripts and using a local binary job script would be enough to indicate whether or not this would actually work.

dirk-thomas: > I'm not sure that this will work.

Please try this PR together with the one on `ament_tools`. It installs the scripts into the libexec path as desired. Since that works I am pretty confident the Debian rules file is able to achieve the same behavior by invoking the same command.

nuclearsandwich: > this PR together with the one on ament_tool

ament tools isn't part of the debian package build. Can you explain why the PR to ament_tools would be needed to validate a debian build with this change?

With dh_python responsible for the direct packaging actions, we don't actually get to invoke setup.py directly. Instead, arguments are passed through the exported environment variables. [[1]](https://github.com/ros2/bloom/blob/b5d17cbbc9dbee3f00f5948ec47ef118fc9cec1e/bloom/generators/debian/generator.py#L539-L543).

The link above is to the ROS 2 bloom fork, the in-progress refactor uses separate templates for the different build types but interaction with setuptools via pybuild environment variables is still used. The ROS 2 bloom fork initially installed python packages into the default debian filesystem hierarchy, after some local testing, https://github.com/ros2/bloom/commit/9969bc18b0890324884e3763c864ace860f26945 and a typo fix for it, https://github.com/ros2/bloom/commit/22e272014ed96e720f74b6363914bb80dd54396d were added. In local testing the prefix alone wasn't sufficient to install python library files in the ROS installation prefix.

We added the install scripts together while I was in CA with https://github.com/ros2/bloom/commit/31fee87b20f6dfbbcf01aedc7776f8f50c64eab4 and its escape fixes https://github.com/ros2/bloom/commit/f221ee781754f661b03e9baec7cbc019d1cc0ce7 because the setting of `--prefix` alone wasn't enough to put the non-libexec binaries in the ROS installation prefix.


Unless the reason prefix was not being honored is identified and resolved for Bloom merging this as-is will require maintainers who want their scripts installed as libexec rather than bin will need to modify the bloom generated templates.

dirk-thomas: > ament tools isn't part of the debian package build. Can you explain why the PR to ament_tools would be needed to validate a debian build with this change?

I wasn't saying that. I just wanted to point out that you can take a closer look to how `ament_tools` builds Python packages. Since it works in that case the Debian rules file should be able to replicate the same invocations and achieve the same behavior.

> We added the install scripts together while I was in CA with ros2/bloom@31fee87 and its escape fixes ros2/bloom@f221ee7 because the setting of `--prefix` alone wasn't enough to put the non-libexec binaries in the ROS installation prefix.

Please try a local build with the modified `ament_tools` package. It doesn't pass the `--install-scripts` argument explicitly yet all executables are installed into the right location (at least for me). If you see a different behavior we need to debug *why* that is the case. But if that works for you too it should be possible to also achieve the same behavior from the Debian rule without (without the need to pass that argument).

mikaelarguedas:
@nuclearsandwich can you try updating the rules to test the approach suggested in these PRs and we will then propagate and get rid of ament_python dependency in the codebase

nuclearsandwich:
> @nuclearsandwich can you try updating the rules to test the approach suggested in these PRs and we will then propagate and get rid of ament_python dependency in the codebase

Yes, as it happens I just finished setting up the reproduction. To try it yourself:

1. Set up the ros2 ros_buildfarm scripts in a virtualenv https://github.com/ros2/ros_buildfarm

2. Create a workspace directory like `mkdir /tmp/binarydeb-python-repro`

3. Run 
```
generate_release_script.py https://raw.githubusercontent.com/nuclearsandwich/build-python-repro-testing/master/buildfarm_config.yaml r2b2 default demo_nodes_py ubuntu xenial amd64 > /tmp/binarydeb-python-repro/run.sh
```
4. Run `cd /tmp/binarydeb-python-repro && sh run.sh`

5. When the job finishes, you can find that dh_python uses pybuild to invoke `setup.py install` with arguments provided by pybuild as well as the `--prefix` argument from https://github.com/nuclearsandwich/test-ros2-demos-release-python-repro/blob/481c3d1bdf248b0ad3192dc9ffc5be3014145bd3/debian/rules#L22 which is modified to specify only `--prefix` and not `--install-scripts` and `--install-lib` as were used for the ROS 2 beta to support the desired python installation behavior.
```
.I: pybuild base:184: /usr/bin/python3 setup.py install --root /tmp/binarydeb/ros-r2b2-demo-nodes-py-0.0.2/debian/ros-r2b2-demo-nodes-py --prefix "/opt/ros/r2b2"
```

6. You can inspect the contents of the package and find that the binaries are installed into the ROS installation prefix thanks to the inserted --install-scripts argument but the other files are installed into /usr.
```
]➤ dpkg -c binary/binarydeb/ros-r2b2-demo-nodes-py_0.0.2-4xenial-20170809-134740-0500_amd64.deb 
drwxr-xr-x root/root         0 2017-08-09 14:50 ./
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/lib/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/
-rwxr-xr-x root/root       344 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/add_two_ints_server
-rwxr-xr-x root/root       322 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/listener
-rwxr-xr-x root/root       356 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/add_two_ints_client_async
-rwxr-xr-x root/root       326 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/talker_qos
-rwxr-xr-x root/root       344 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/add_two_ints_client
-rwxr-xr-x root/root       330 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/listener_qos
-rwxr-xr-x root/root       318 2017-08-09 14:50 ./opt/ros/r2b2/lib/demo_nodes_py/talker
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/share/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/share/demo_nodes_py/
-rw-r--r-- root/root       878 2017-08-09 14:47 ./opt/ros/r2b2/share/demo_nodes_py/package.xml
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/share/ament_index/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/share/ament_index/resource_index/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./opt/ros/r2b2/share/ament_index/resource_index/packages/
-rw-r--r-- root/root         0 2017-06-30 21:30 ./opt/ros/r2b2/share/ament_index/resource_index/packages/demo_nodes_py
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/lib/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/lib/python3/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/lib/python3/dist-packages/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/lib/python3/dist-packages/demo_nodes_py-0.0.0.egg-info/
-rw-r--r-- root/root       538 2017-08-09 14:50 ./usr/lib/python3/dist-packages/demo_nodes_py-0.0.0.egg-info/PKG-INFO
-rw-r--r-- root/root       433 2017-08-09 14:50 ./usr/lib/python3/dist-packages/demo_nodes_py-0.0.0.egg-info/entry_points.txt
-rw-r--r-- root/root         0 2017-08-09 14:50 ./usr/lib/python3/dist-packages/demo_nodes_py-0.0.0.egg-info/requires.txt
-rw-r--r-- root/root         1 2017-08-09 14:50 ./usr/lib/python3/dist-packages/demo_nodes_py-0.0.0.egg-info/dependency_links.txt
-rw-r--r-- root/root         1 2017-08-09 14:50 ./usr/lib/python3/dist-packages/demo_nodes_py-0.0.0.egg-info/top_level.txt
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/share/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/share/doc/
drwxr-xr-x root/root         0 2017-08-09 14:50 ./usr/share/doc/ros-r2b2-demo-nodes-py/
-rw-r--r-- root/root       257 2017-08-09 14:47 ./usr/share/doc/ros-r2b2-demo-nodes-py/changelog.Debian.gz
```

I understand the behavior using only `--prefix` in environments ament_tools is generally used in is sufficient. But it is not sufficient in the environment our debian packages are currently being built under. This could be because of some issue with the debian control or rules templates that we've created for ROS 2 or it could be a situation somewhat similar to https://github.com/ros-infrastructure/bloom/issues/327 where the Debian tooling has implicit defaults which must be explicitly overridden.

When I investigated this issue previously, I found that the explicit `--install-scripts` and `--install-libs` were both necessary for the desired behavior.

dirk-thomas:
@nuclearsandwich Any update on this?

nuclearsandwich:
> Your above example proves that the executables are installed into the libexec path as desired.

Yes, but this was never in dispute. The precise issue I am raising is that we cannot remove the explicit `--install-scripts` argument from the bloom templates as you have done for ament tools and maintain the desired behavior.

> I didn't mention --install-libs anywhere. Please keep it as-is for this comparison since that option is not related to the script location

It's not related to the script location but it is absolutely related to the feedback I've been providing. Specifically that the assertion (made here: https://github.com/ros2/demos/pull/158#discussion_r131944588)  that `--prefix` is the only necessary argument when building debian packages 1. makes intuitive sense and seems like it should be the case and 2. is demonstrably false as indicated by the necessity of the Bloom changes I pointed to above, particularly https://github.com/ros2/bloom/commit/9969bc18b0890324884e3763c864ace860f26945.

The reproduction for demo_nodes_py show in a concise fashion that bloom changes you would mandate in this PR do not create a correct package. When either of `--install-libs` or `--install-scripts` is missing from the debian configuration and none is inserted at runtime, parts of the package will not be installed correctly.

For further clarification, I've created a second demonstration of why changes requested in bloom for this PR to be sufficient do not work. Below is a list of package contents for ros2cli with the changes to bloom that would be required to support this PR as-is. You can reproduce it following the steps above substituting the package name `ros2cli` for `demo_nodes_py`.

Here are the package contents:
```
➤ dpkg -c binary/binarydeb/ros-r2b2-ros2cli_0.0.2-2xenial-20170810-135140-0500_amd64.deb 
drwxr-xr-x root/root         0 2017-08-10 14:53 ./
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli-0.0.0.egg-info/
-rw-r--r-- root/root       638 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli-0.0.0.egg-info/PKG-INFO
-rw-r--r-- root/root       565 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli-0.0.0.egg-info/entry_points.txt
-rw-r--r-- root/root        26 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli-0.0.0.egg-info/requires.txt
-rw-r--r-- root/root         1 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli-0.0.0.egg-info/dependency_links.txt
-rw-r--r-- root/root         8 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli-0.0.0.egg-info/top_level.txt
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/daemon/
-rw-r--r-- root/root      4327 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/daemon/__init__.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/daemon/__pycache__/
-rw-r--r-- root/root      3818 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/daemon/__pycache__/__init__.cpython-35.pyc
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/
-rw-r--r-- root/root      2793 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/extensions.py
-rw-r--r-- root/root      4096 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/__init__.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/__pycache__/
-rw-r--r-- root/root      2009 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/__pycache__/extension_points.cpython-35.pyc
-rw-r--r-- root/root      3537 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/__pycache__/__init__.cpython-35.pyc
-rw-r--r-- root/root      2059 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/__pycache__/extensions.cpython-35.pyc
-rw-r--r-- root/root       958 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/__pycache__/daemon.cpython-35.pyc
-rw-r--r-- root/root      2613 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/extension_points.py
-rw-r--r-- root/root      1357 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/command/daemon.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/
-rw-r--r-- root/root         0 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/__init__.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/__pycache__/
-rw-r--r-- root/root      1879 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/__pycache__/direct.cpython-35.pyc
-rw-r--r-- root/root      1459 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/__pycache__/strategy.cpython-35.pyc
-rw-r--r-- root/root       149 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/__pycache__/__init__.cpython-35.pyc
-rw-r--r-- root/root      3145 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/__pycache__/daemon.cpython-35.pyc
-rw-r--r-- root/root      1531 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/strategy.py
-rw-r--r-- root/root      1871 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/direct.py
-rw-r--r-- root/root      3407 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/node/daemon.py
-rw-r--r-- root/root      3834 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/plugin_system.py
-rw-r--r-- root/root         0 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/__init__.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/__pycache__/
-rw-r--r-- root/root      2436 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/__pycache__/entry_points.cpython-35.pyc
-rw-r--r-- root/root       144 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/__pycache__/__init__.cpython-35.pyc
-rw-r--r-- root/root      3223 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/__pycache__/plugin_system.cpython-35.pyc
-rw-r--r-- root/root      1344 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/__pycache__/cli.cpython-35.pyc
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/
-rw-r--r-- root/root      1060 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/start.py
-rw-r--r-- root/root      1339 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/__init__.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/__pycache__/
-rw-r--r-- root/root       800 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/__pycache__/start.cpython-35.pyc
-rw-r--r-- root/root      1304 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/__pycache__/__init__.cpython-35.pyc
-rw-r--r-- root/root       729 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/__pycache__/status.cpython-35.pyc
-rw-r--r-- root/root       840 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/__pycache__/stop.cpython-35.pyc
-rw-r--r-- root/root      1104 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/stop.py
-rw-r--r-- root/root       972 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/daemon/status.py
-rw-r--r-- root/root      1786 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/__init__.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/__pycache__/
-rw-r--r-- root/root      1588 2017-08-10 14:53 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/verb/__pycache__/__init__.cpython-35.pyc
-rw-r--r-- root/root      2317 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/cli.py
-rw-r--r-- root/root      3378 2017-07-01 02:34 ./opt/ros/r2b2/lib/python3.5/site-packages/ros2cli/entry_points.py
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/share/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/share/ros2cli/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./opt/ros/r2b2/share/ros2cli/environment/
-rw-r--r-- root/root       815 2017-07-01 02:34 ./opt/ros/r2b2/share/ros2cli/environment/ros2-argcomplete.zsh
-rw-r--r-- root/root       817 2017-07-01 02:34 ./opt/ros/r2b2/share/ros2cli/environment/ros2-argcomplete.bash
drwxr-xr-x root/root         0 2017-08-10 14:53 ./usr/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./usr/bin/
-rwxr-xr-x root/root       296 2017-08-10 14:53 ./usr/bin/ros2
-rwxr-xr-x root/root       312 2017-08-10 14:53 ./usr/bin/_ros2_daemon
drwxr-xr-x root/root         0 2017-08-10 14:53 ./usr/share/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./usr/share/doc/
drwxr-xr-x root/root         0 2017-08-10 14:53 ./usr/share/doc/ros-r2b2-ros2cli/
-rw-r--r-- root/root       251 2017-08-10 14:51 ./usr/share/doc/ros-r2b2-ros2cli/changelog.Debian.gz
```

With only `--prefix` and `--install-lib` and without `--install-scripts` the executables `ros2` and `_ros2_daemon` are installed into `/usr/bin`

This PR currently will not replace an earlier `--install-scripts` argument. Per https://github.com/ros2/demos/pull/158#discussion_r131944588 this is an intended break from the existing behavior in ament_python in a way that is incompatible with debian package creation for ROS 2.

dirk-thomas:
> I've created a second demonstration

Can you please share the steps how to reproduce this. Thanks.

nuclearsandwich:
> Can you please share the steps how to reproduce this. Thanks.

> > You can reproduce it following the steps above substituting the package name ros2cli for demo_nodes_py.

dirk-thomas:
The missing piece of information was in `dhpython` - specifically `build/plugin_distutils.py:33-34`:

> distutils doesn't have sane command-line API - this decorator creates
> .pydistutils.cfg file to workaround it

The generated `.pydistutils.cfg` file to configure the build contains the following:

```
[clean]
all=1
[build]
build-lib=/tmp/binarydeb/ros-r2b2-ros2cli-0.0.2/.pybuild/pythonX.Y_3.5/build
[install]
install-layout=deb
install-scripts=/usr/bin
install-lib=/usr/lib/python3.5/dist-packages
[easy_install]
allow_hosts=None
```

From that file only `build-lib` and `install-lib` are configurable by arguments. The content of `install-scripts` is hard coded.

Changing the cfg generated by `dhpython` doesn't seem like a good idea since anybody trying to release the package on their own would have the same problem. Making the logic in the `setup.py` file more complicated doesn't seem like an option either - it was already pretty complicates as is.

Instead I updated this PR as well as ros2/examples#178 to configure the libexec location for the scripts in a package specific `setup.cfg` files. @nuclearsandwich Please try if the current state works for the release process (it worked for me locally).

nuclearsandwich:
> Instead I updated this PR as well as ros2/examples#178 to configure the libexec location for the scripts in a package specific setup.cfg files. @nuclearsandwich Please try if the current state works for the release process (it worked for me locally).

I tested this in two configurations. To reproduce them you can fork https://github.com/nuclearsandwich/build-python-repro-testing and set the revision of demo_nodes_py in rosdistro.yaml and r2b2-rosdistro-cache.yaml.

`0.2.0-6`: With the current `ros2` behavior (setting `--prefix`, `--install-lib`, and `--install-scripts` in debian/rules):

- scripts installed into the rosdistro prefix bin
- library files installed into the rosdistro prefix


`0.2.0-7`: With only `--prefix` set in debian/rules:

- scripts installed into libexec
- library files installed into /usr

Using a setup.cfg file is definitely a clearer, preferable solution to inserting command line arguments. The setup.cfg is still overridden by the explicit command line arguments passed via the current Bloom debian/rules template. In order to honor the setup.cfg we'll need to remove `--install-scripts` from the template and add `setup.cfg` files for all other python packages whether they install into bin or libexec.

I am also strongly (like 9/10) in favor of adding install-lib to the setup.cfg as well. It keeps the template magic to a minimum and keeps the prefix-relative setup options together. Since every python package needs a setup.cfg whether it is installing scripts to bin or libexec, it doesn't increase the number of packages we need to modify to enact this change.

dirk-thomas:
I don't think a package developer should need to specify `install-lib` in the setup.cfg. The location of the scripts is different since the goal is to customize the default (base-relative) location. The location of lthe libraries is imo a packaging decision and should therefore stay in the rules file.

dirk-thomas:
> Since every python package needs a setup.cfg whether it is installing scripts to bin or libexec, it doesn't increase the number of packages we need to modify to enact this change.

No package has a setup.cfg file at the moment. And only the ones which customize the script install location needs one.

nuclearsandwich:
> No package has a setup.cfg file at the moment. And only the ones which customize the script install location need one.

Can you show your work here? What combination of debian/rules template and setup.cfg allows packages installing into the ROS prefix bin not to need a setup.cfg?

dirk-thomas:
> Can you show your work here? What combination of debian/rules template and setup.cfg allows packages installing into the ROS prefix bin not to need a setup.cfg?

The current set of PRs together with the removal of `--install-scripts` from the bloom template work for me.

Only the packages which want to install scripts into libexec have a `setup.cfg` file. And the bloom template keeps passing `--prefix` as well as `--install-lib`.

[`--install-lib` should use `dist-packages` instead of `sire-packages` since we want the Debian specific layout but that is an unrelated change once the patches to `bloom` are getting reviewed.]

nuclearsandwich:
> The current set of PRs together with the removal of --install-scripts from the bloom template work for me.

Are you proposing we remove --install-scripts from the bloom templates in source or patch them out for packages using libexec. Removing `--install-scripts` from bloom altogether will cause packages with no setup.cfg to install scripts to /usr/bin because of the generated `.pydistutils.cfg`. Leaving `--install-scripts` for all packages, including libexec ones overrides the setup.cfg values and still places scripts in the ROS prefix bin (see https://github.com/ros2/demos/pull/158#issuecomment-322478223 for repro info and my results)

dirk-thomas:
> Removing --install-scripts from bloom altogether will cause packages with no setup.cfg to install scripts to /usr/bin because of the generated .pydistutils.cfg. Leaving --install-scripts for all packages, including libexec ones overrides the setup.cfg values and still places scripts in the ROS prefix bin

You are right. So the `bloom` template should use `--install-scripts` only when the package doesn't override that location (is have a `setup.cfg` with a `[install]` section and a `install-scripts` key).

nuclearsandwich:
> So the bloom template should use --install-scripts only when the package doesn't override that location (the package having a setup.cfg with a [install] section and a install-scripts key).

To my knowledge or @wjwwood's bloom doesn't currently look at package contents beyond the package.xml and I am not sure from a hygiene / design perspective whether we really want it testing file presence let alone parsing other package files.

Everything else bloom needs comes from the package.xml. Could we put something there to indicate either 1. where to install scripts (defaulting to bin) or 2. to leave options off and trust that a setup.cfg is present.

dirk-thomas:
> Could we put something there to indicate either 1. where to install scripts (defaulting to bin)

I don't think this information belongs into the manifest.

> 2. to leave options off and trust that a setup.cfg is present.

That would require every package to have a `setup.cfg` file even though it is only required during packaging. I would rather not force that effort onto every package maintainer.

mikaelarguedas:
Offline discussion lead to the conclusion that we will have bloom intelligently add the scripts argument if there is no `setup.cfg` or if there is a `setup.cfg` but it does not set the scripts directory. If bloom sets the scripts directory it will be `$base/bin`. We can:

- add a default `setup.cfg` in the new python package template which installs executables to libexec
- create a wiki page that summarizes this issue
- talk to @j-rivero about getting debhelper to use `$base` which would remove the need for the custom logic in bloom

We considered requiring all python packages have a `setup.cfg` which sets the scripts directory in order to release, but that would require us to create and inject a `setup.cfg` for python packages which did not have one upstream, e.g. if we wanted to release the flake8 python package on our farm for some reason it would not have a `setup.cfg` most likely and we would have to add one, which seemed like a bad requirement. It would make us just a little bit less standard in our process. Also we would still need to introspect the `setup.cfg` if we wanted to give a decent warning to the user when they have one but failed to set a scripts directory.

dirk-thomas:
Based on the conclusion it would be great if someone can review ament/ament_python#3 as well as all connected PRs so that they can get merged. The changes to `bloom` can be applied anytime between now and before the Beta 3.

topic_monitor/setup.py
```diff
@@ -1,17 +1,17 @@
...
 from setuptools import find_packages
 from setuptools import setup
 
 package_name = 'topic_monitor'
-data_files = get_data_files(package_name)
-install_scripts_to_libexec(package_name)
 
 setup(
-    name='topic_monitor',
+    name=package_name,
```
mikaelarguedas: nice catch!

PR #161, opened 2017-08-11
Use wait_for_service in parameter demos

dhood:
connects to ros2/rclcpp#356

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2993/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/424/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2419/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3094/)

fixes test flakiness listed in https://github.com/ros2/build_cop/issues/41


demo_nodes_cpp/src/parameters/list_parameters.cpp
```diff
@@ -27,7 +29,15 @@ int main(int argc, char ** argv)
   auto parameter_service = std::make_shared<rclcpp::parameter_service::ParameterService>(node);
 
   auto parameters_client = std::make_shared<rclcpp::parameter_client::SyncParametersClient>(node);
+  while (!parameters_client->wait_for_service(1s)) {
+    if (!rclcpp::ok()) {
+      std::cout << "Interrupted while waiting for the service. Exiting." << std::endl;
```
dirk-thomas: Use `printf`.

Same below.

mikaelarguedas: This entire file uses streams. If changing to printf, all prints should be updated. This can be done either here or in a separate PR

dirk-thomas: I am not saying that the existing `cout`'s need to be updated in this PR (though since it would only take a couple of minutes I think it would be good to do).

But newly added code should not use it.

demo_nodes_cpp/src/parameters/parameter_node.cpp
```diff
@@ -47,11 +47,20 @@ int main(int argc, char ** argv)
...
+    }
+    printf("service not available, waiting again...\n");
+  }
 
   // Setup callback for changes to parameters.
   auto sub = parameters_client->on_parameter_event(on_parameter_event);
 
   rclcpp::spin(node);
 
+  rclcpp::shutdown();
```
dirk-thomas: Should the other executables also call `shutdown`?

dhood: done in https://github.com/ros2/demos/pull/162

PR #162, opened 2017-08-11
Call to rclcpp::shutdown in demos so rclcpp signal handler gets removed

dhood:
connects to ros2/rmw_implementation#25

combined with https://github.com/ros2/rclcpp/pull/353 this fixes https://github.com/ros2/rmw_implementation/issues/25 (also listed in https://github.com/ros2/build_cop/issues/41)

Calling to shutdown is required because otherwise if launch_testing triggers an interrupt while the process is exiting the process deadlocks when multiple rmw implementions are supported (and therefore poco is in use). see https://github.com/ros2/rmw_implementation/issues/25

### DIFF
--- a/composition/src/api_composition.cpp
+++ b/composition/src/api_composition.cpp
@@ -159,5 +159,6 @@ int main(int argc, char * argv[])
     exec.remove_node(node);
   }
   nodes.clear();
+  rclcpp::shutdown();
   return 0;
 }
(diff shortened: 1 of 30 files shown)


composition/src/api_composition.cpp
```diff
@@ -159,5 +159,6 @@ int main(int argc, char * argv[])
     exec.remove_node(node);
   }
   nodes.clear();
+  rclcpp::shutdown();
```
mikaelarguedas: Nit: no empty line before or after while you add one in the other files of this PR

PR #167, opened 2017-08-30
match xml schema

mikaelarguedas:


mikaelarguedas:
>  It would probably be best to add a linter for this if we care about schema compliance.

Yeah I don't know if we care enough to enforce it automatically, I'd say its a nice-to-have. I just fixed those because I had xmllint running on both my ros1 and ros2 workspaces when investigating https://github.com/ros/rospack/issues/77 and that this was the only non-forked ROS 2 package not complying

PR #169, opened 2017-09-07
Use features from executors pull request

sloretz:
This updates the demo code to inherit from rclpy.Node and use rclpy.spin().

connects to ros2/rclpy#104

### DIFF
--- a/demo_nodes_py/demo_nodes_py/services/add_two_ints_server.py
+++ b/demo_nodes_py/demo_nodes_py/services/add_two_ints_server.py
@@ -17,26 +17,29 @@ from example_interfaces.srv import AddTwoInts
 import rclpy
 
 
-def add_two_ints_callback(request, response):
-    response.sum = request.a + request.b
-    print('Incoming request\na: %d b:%d' % (request.a, request.b))
+class AddTwoIntsServer(rclpy.Node):
 
-    return response
+    def __init__(self):
+        super().__init__('add_two_ints_server')
+        self.srv = self.create_service(AddTwoInts, 'add_two_ints', self.add_two_ints_callback)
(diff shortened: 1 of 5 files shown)


demo_nodes_py/demo_nodes_py/services/add_two_ints_server.py
```diff
@@ -17,26 +17,29 @@
...
-def add_two_ints_callback(request, response):
-    response.sum = request.a + request.b
-    print('Incoming request\na: %d b:%d' % (request.a, request.b))
+class AddTwoIntsServer(rclpy.Node):
 
-    return response
+    def __init__(self):
+        super().__init__('add_two_ints_server')
+        self.srv = self.create_service(AddTwoInts, 'add_two_ints', self.add_two_ints_callback)
+        assert self.srv  # prevent unused variable warning
```
dirk-thomas: This line does more than prevent the warning but also ensure that the function returned a non-empty value. So I think the comment is a bit misleading. Therefore I would either suggest to not store the returned value if it is never user or use a different statement (e.g. just `self.srv`) to suppress the warning (I haven't tried that but I think that should do it).

Same below.

sloretz:
PR squashed/merged. Will address issues pointed out by @dirk-thomas in a PR to be merged after r2b3.

demo_nodes_py/demo_nodes_py/services/add_two_ints_server.py
sloretz: I removed the line in 6f80c163302df193850781a3e6eb991e807ac136. The comment was from before it was a member variable. I don't see the warning in the tests after removing the line.

PR #176, opened 2017-09-15
Follow up to executor comments

sloretz:
Follow up to address comments from #169 that didn't get addressed before r2b3

connects to ros2/examples#184


sloretz:
Builds passed, and approved. Merging

PR #181, opened 2017-10-04
lint generated launch files

dirk-thomas:
Follow up of https://github.com/ros2/demos/pull/157#issuecomment-320448804

This patch shows how to lint generated files. It does so only for the composition package. Other packages should probably use something similar.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3305/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/572/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2637/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3344/)

### DIFF
--- a/composition/CMakeLists.txt
+++ b/composition/CMakeLists.txt
@@ -154,6 +154,7 @@ if(BUILD_TESTING)
     "test_ament_index/$<CONFIG>/share/ament_index/resource_index/node_plugin/${PROJECT_NAME}"
     CONTENT "${node_plugins}")
 
+  set(generated_python_files)
   macro(tests)
     set(MANUAL_COMPOSITION_EXECUTABLE $<TARGET_FILE:manual_composition>)
     set(LINKTIME_COMPOSITION_EXECUTABLE $<TARGET_FILE:linktime_composition>)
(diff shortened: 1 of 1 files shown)


composition/CMakeLists.txt
```diff
@@ -190,6 +194,13 @@ if(BUILD_TESTING)
   endif()
 
   call_for_each_rmw_implementation(tests)
+
+  find_package(ament_cmake_flake8 REQUIRED)
```
mikaelarguedas: is this necessary ? this package already find_package the `ament_lint_auto` dependencies.
If we want the ability to make some of them `REQUIRED` we could implement something similar to https://github.com/ament/ament_cmake/pull/93 in ament_lint_auto and just pass ament_cmake_flake8 as a required dependency

dirk-thomas: I am just applying "find what you need". Since the code is using a CMake function of that package it is finding it explicitly.

dirk-thomas:
Waiting for review.

PR #182, opened 2017-10-13
Replace time.sleep() with wait_for_service()

sloretz:
Changes demos to use wait_for_service()

### DIFF
--- a/demo_nodes_py/demo_nodes_py/services/add_two_ints_client.py
+++ b/demo_nodes_py/demo_nodes_py/services/add_two_ints_client.py
@@ -24,7 +24,7 @@ def main(args=None):
 
     cli = node.create_client(AddTwoInts, 'add_two_ints')
     while not cli.wait_for_service(timeout_sec=1.0):
-        print("Still waiting for service")
+        print("service not available, waiting again...")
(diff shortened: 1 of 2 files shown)


sloretz:
@wjwwood Added loop and print() in 5cc0dc4b63bc20b2b2baf6a625af25223b145163

mikaelarguedas:
I don't think we check this console output anywhere, but it may be useful in the future to print the same string as in [c++](https://github.com/ros2/demos/blob/082f5b30e2e4ac4db440594d2bc1782def255516/demo_nodes_cpp/src/services/add_two_ints_client.cpp#L79)

sloretz:
@mikaelarguedas Changed the message to "service not available, waiting again..." 2cdce5860469b2ad07d40bd22902f8b00c1ebc14 5805fa040e47165d32c4ec4de3d6e1b94bcd549b

sloretz:
I'll think this can be merged now that ros2/rclpy#161 has been merged.

mikaelarguedas:
@sloretz sorry I missed that this was in review, would you mind doing a follow up PR to use the loggers rather than `print` like the other python demos?

PR #183, opened 2017-10-17
add library_path env hook

dirk-thomas:
Fixes ament/ament_cmake#111.

### DIFF
--- a/composition/CMakeLists.txt
+++ b/composition/CMakeLists.txt
@@ -73,6 +73,8 @@ ament_target_dependencies(client_component
 rclcpp_register_node_plugins(client_component "composition::Client")
 set(node_plugins "${node_plugins}composition::Client;$<TARGET_FILE:client_component>\n")
 
+# since the package installs libraries without exporting them
+# it needs to make sure that the library path is being exported
 if(NOT WIN32)
   ament_environment_hooks(
     "${ament_cmake_package_templates_ENVIRONMENT_HOOK_LIBRARY_PATH}")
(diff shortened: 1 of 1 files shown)


composition/CMakeLists.txt
```diff
@@ -73,6 +73,11 @@ ament_target_dependencies(client_component
 rclcpp_register_node_plugins(client_component "composition::Client")
 set(node_plugins "${node_plugins}composition::Client;$<TARGET_FILE:client_component>\n")
 
+if(NOT WIN32)
+  ament_environment_hooks(
+    "${ament_cmake_package_templates_ENVIRONMENT_HOOK_LIBRARY_PATH}")
```
wjwwood: This code isn't very self explanatory, and it touches internal variables of `ament_cmake` directly. I think it would be better to add a new function or macro to do this (which can be documented and is searchable) and call it from here and from the `ament_export_library` call. You could call it `ament_add_library_path_hook()` or something like that.

In either case (new function or leaving this as-is), a comment here explaining why this needs to be called explicitly is needed in my opinion.

dirk-thomas: > it touches internal variables of ament_cmake directly.

The variable is not meant to be internal.

> I think it would be better to add a new function or macro to do this

The function `ament_environment_hooks` is documented and it takes arbitrary env hooks. I don't think it makes sense to create functions for *every* possible parameter.

> a comment here explaining why this needs to be called explicitly is needed in my opinion.

Absolutely, done in 5760ea4da62d3edeec35604a3ce48a29ce883119.

dirk-thomas: See ament/ament_cmake#112 about a generic comment which variables are being generated.

PR #185, opened 2017-10-23
Demo nodes for raw publish and subscribe 

Karsten1987:
I am making this the top level PR to gather all related PRs. This is WIP but as it affects basically all packages down the stack, I open this PR for early visibility.

This PR (and all others) aim to expose the raw CDR stream through the ROS2 API. This enables to send a raw data stream directly to the wire and in return take the data without going through the serialization process. 
This ultimately serves as the base for a future upcoming ROSbag implementation.

### DIFF
--- a/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
+++ b/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
@@ -169,7 +169,7 @@ int main(int argc, char ** argv)
     rclcpp::message_memory_strategy::MessageMemoryStrategy<
       std_msgs::msg::UInt32, MyAllocator<>>>(alloc);
   auto subscriber = node->create_subscription<std_msgs::msg::UInt32>(
-    "allocator_tutorial", 10, callback, nullptr, false, msg_mem_strat, alloc);
+    "allocator_tutorial", callback, 10, nullptr, false, msg_mem_strat, alloc);
 
   // Create a MemoryStrategy, which handles the allocations made by the Executor during the
   // execution path, and inject the MemoryStrategy into the Executor.
--- a/demo_nodes_cpp/src/topics/listener_raw.cpp
+++ b/demo_nodes_cpp/src/topics/listener_raw.cpp
@@ -1,0 +1,69 @@
+// Copyright 2014 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+#include <memory>
+#include <string>
+
+#include "rcl/types.h"
+#include "rclcpp/rclcpp.hpp"
+#include "rcutils/cmdline_parser.h"
+
+#include "std_msgs/msg/string.hpp"
+#include "test_msgs/msg/nested.hpp"
+
+void print_usage()
+{
+  printf("Usage for listener app:\n");
+  printf("listener [-t topic_name] [-h]\n");
+  printf("options:\n");
+  printf("-h : Print this help function.\n");
+  printf("-t topic_name : Specify the topic on which to subscribe. Defaults to chatter.\n");
+}
+
+void chatterCallback_raw(const std::shared_ptr<rcl_message_raw_t> msg)
+{
+  std::cout << "I heard: [" << msg->buffer_length << "]" << std::endl;
+  for (size_t i = 0; i < msg->buffer_length; ++i) {
+    fprintf(stderr, "%02x ", msg->buffer[i]);
+  }
+  fprintf(stderr, "\n");
+}
+
+void chatterCallback(const std::shared_ptr<test_msgs::msg::Nested> msg)
+{
+  std::cout << "I heard: [" << msg->primitive_values.int32_value << "]" << std::endl;
+}
+
+int main(int argc, char * argv[])
+{
+  rclcpp::init(argc, argv);
+  auto node = rclcpp::Node::make_shared("listener");
+
+  if (rcutils_cli_option_exist(argv, argv + argc, "-h")) {
+    print_usage();
+    return 0;
+  }
+
+  auto topic = std::string("chatter");
+  if (rcutils_cli_option_exist(argv, argv + argc, "-t")) {
+    topic = std::string(rcutils_cli_get_option(argv, argv + argc, "-t"));
+  }
+  auto sub = node->create_subscription<test_msgs::msg::Nested>(
+    topic, chatterCallback_raw);
+
+  rclcpp::spin(node);
+
+  return 0;
+}
--- a/demo_nodes_cpp/src/topics/listener_raw.cpp
+++ b/demo_nodes_cpp/src/topics/listener_raw.cpp
@@ -21,7 +21,6 @@
 #include "rcutils/cmdline_parser.h"
 
 #include "std_msgs/msg/string.hpp"
-#include "test_msgs/msg/nested.hpp"
 
 void print_usage()
 {
@@ -41,11 +40,6 @@ void chatterCallback_raw(const std::shared_ptr<rcl_message_raw_t> msg)
   fprintf(stderr, "\n");
 }
 
-void chatterCallback(const std::shared_ptr<test_msgs::msg::Nested> msg)
-{
-  std::cout << "I heard: [" << msg->primitive_values.int32_value << "]" << std::endl;
-}
-
 int main(int argc, char * argv[])
 {
   rclcpp::init(argc, argv);
@@ -60,7 +54,7 @@ int main(int argc, char * argv[])
   if (rcutils_cli_option_exist(argv, argv + argc, "-t")) {
     topic = std::string(rcutils_cli_get_option(argv, argv + argc, "-t"));
   }
-  auto sub = node->create_subscription<test_msgs::msg::Nested>(
+  auto sub = node->create_subscription<std_msgs::msg::String>(
     topic, chatterCallback_raw);
 
   rclcpp::spin(node);
--- a/demo_nodes_cpp/src/topics/listener_raw.cpp
+++ b/demo_nodes_cpp/src/topics/listener_raw.cpp
@@ -31,33 +31,65 @@ void print_usage()
   printf("-t topic_name : Specify the topic on which to subscribe. Defaults to chatter.\n");
 }
 
-void chatterCallback_raw(const std::shared_ptr<rcl_message_raw_t> msg)
+// Create a Listener class that subclasses the generic rclcpp::Node base class.
+// The main function below will instantiate the class as a ROS node.
+class RawListener : public rclcpp::Node
 {
-  std::cout << "I heard: [" << msg->buffer_length << "]" << std::endl;
-  for (size_t i = 0; i < msg->buffer_length; ++i) {
-    fprintf(stderr, "%02x ", msg->buffer[i]);
+public:
+  explicit RawListener(const std::string & topic_name)
+  : Node("raw_listener")
+  {
+    // Create a callback function for when messages are received.
+    // Variations of this function also exist using, for example UniquePtr for zero-copy transport.
+    auto callback =
+      [this](const std::shared_ptr<rcl_message_raw_t> msg) -> void
+      {
+        std::cout << "I heard: [" << msg->buffer_length << "]" << std::endl;
+        for (size_t i = 0; i < msg->buffer_length; ++i) {
+          printf("%02x ", msg->buffer[i]);
+        }
+        printf("\n");
+      };
+
+    // Create a subscription to the topic which can be matched with one or more compatible ROS
+    // publishers.
+    // Note that not all publishers on the same topic with the same type will be compatible:
+    // they must have compatible Quality of Service policies.
+    sub_ = create_subscription<std_msgs::msg::String>(topic_name, callback);
   }
-  fprintf(stderr, "\n");
-}
+
+private:
+  rclcpp::Subscription<rmw_message_raw_t>::SharedPtr sub_;
+};
 
 int main(int argc, char * argv[])
 {
-  rclcpp::init(argc, argv);
-  auto node = rclcpp::Node::make_shared("listener");
+  // Force flush of the stdout buffer.
+  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
 
   if (rcutils_cli_option_exist(argv, argv + argc, "-h")) {
     print_usage();
     return 0;
   }
 
+  // Initialize any global resources needed by the middleware and the client library.
+  // You must call this before using any other part of the ROS system.
+  // This should be called once per process.
+  rclcpp::init(argc, argv);
+
+  // Parse the command line options.
   auto topic = std::string("chatter");
   if (rcutils_cli_option_exist(argv, argv + argc, "-t")) {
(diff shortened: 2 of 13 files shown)


demo_nodes_cpp/src/topics/talker_raw.cpp
```diff
@@ -0,0 +1,71 @@
...
+  rcl_message_raw_t raw_msg;
+  raw_msg.buffer_length = 24;
+  raw_msg.buffer = new char[raw_msg.buffer_length];
+
+  int i = 1;
+
+  while (rclcpp::ok()) {
+    sprintf(raw_msg.buffer, "%c%c%c%c%c%c%c%c%s %d",
+      0x00, 0x01, 0x00, 0x00, 0x0f, 0x00, 0x00,0x00, "hello world", (++i));
+    std::cout << "Publishing: '" << raw_msg.buffer << "'" << std::endl;
```
dirk-thomas: Use [printf](https://github.com/ros2/ros2/wiki/Developer-Guide#c-specific).

Karsten1987: copied from the existing `talker example`. It's outside of this PR, but I'll update the talker code accordingly.

mikaelarguedas: The other demo code will be updated to use logging pretty soon so I think it's fine to just use printf in the new code without modifying existing one

Karsten1987: @mikaelarguedas Just saw your comment. I'll rebase accordingly if necessary.

demo_nodes_cpp/src/topics/talker.cpp
```diff
@@ -57,7 +57,7 @@ int main(int argc, char * argv[])
 
   while (rclcpp::ok()) {
     msg->data = "Hello World: " + std::to_string(i++);
-    std::cout << "Publishing: '" << msg->data << "'" << std::endl;
+    printf("Publishing: %s\n", msg->data.c_str());
```
mikaelarguedas: If we modify this in this PR, it needs to print the exacte same message as before (with single quotes around the data). Otherwise the tests will fail because they expect them:
https://github.com/ros2/demos/blob/0c2fa6b03b96790262932b854646b3fdf21b1bdf/demo_nodes_cpp/test/talker.txt#L1

Karsten1987:
I think this is up for review.

As I believe this is a quite huge PR, I'll explain the details on each referenced PR if necessary.

CI:
linux: [Build Status](https://ci.ros2.org/job/ci_linux/4027/)
osx: [Build Status](https://ci.ros2.org/job/ci_osx/3354/)
win: [Build Status](https://ci.ros2.org/job/ci_windows/4118/) (unrelated)

demo_nodes_cpp/src/topics/talker_raw.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+class RawTalker : public rclcpp::Node
+{
+public:
+  explicit RawTalker(const std::string & topic_name)
+  : Node("raw_talker")
+  {
+    raw_msg_.buffer_length = 24;
+    raw_msg_.buffer = new char[raw_msg_.buffer_length];
```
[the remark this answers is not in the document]
dirk-thomas: While the `delete` in the destructor is sufficient for the demo code to be "correct" I would suggest to use a smart pointer. Otherwise this might encourage people to copy-n-paste this code which is "incomplete" without a copy constructor and assignment operator.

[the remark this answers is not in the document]
Karsten1987: agreed, a smart pointer would be nice, but I don't think it's more user friendly. I would rather say it's even more error prone to make copy-paste mistakes as you can't simply instantiate a shared ptr with that c-struct. You have to pass a custom deleter function / callable object with the smart pointer to clean up correctly. Does this make sense?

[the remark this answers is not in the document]
dirk-thomas: A `std::unique_ptr` is specialized for arrays and should be fine to use (as of C++11) without a custom deleter afaik (see http://en.cppreference.com/w/cpp/memory/unique_ptr).

wjwwood:
@Karsten1987 Should/can this be put into "in progress"? I have outstanding comments in https://github.com/ros2/rmw_fastrtps/pull/186 and https://github.com/ros2/rmw_connext/pull/259 (at least I think so) and this has stayed in review for several days. It would help me to put it in progress until issues are addressed so that I don't have to reevaluate its state each time I go through waffle to check for needed reviews.

demo_nodes_cpp/src/topics/listener_raw.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    // Create a subscription to the topic which can be matched with one or more compatible ROS
+    // publishers.
+    // Note that not all publishers on the same topic with the same type will be compatible:
+    // they must have compatible Quality of Service policies.
+    sub_ = create_subscription<std_msgs::msg::String>(topic_name, callback);
+  }
+
+private:
+  rclcpp::Subscription<rmw_message_raw_t>::SharedPtr sub_;
```
sloretz: `rcl_message_raw_t` is used in the callback. Maybe keep the type consistent?

demo_nodes_cpp/src/topics/listener_raw.cpp
```diff
@@ -0,0 +1,106 @@
...
+};
+
+int main(int argc, char * argv[])
+{
+  // Force flush of the stdout buffer.
+  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
```
sloretz: Since this is a c++ example what do you think about using `std::cout` instead? The demo could either `std::cout << std::unitbuf` to always flush or `std::cout << std::flush` to explicitly flush.

sloretz: same comment for talker_raw.cpp

demo_nodes_cpp/src/topics/talker_raw.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+public:
+  explicit RawTalker(const std::string & topic_name)
+  : Node("raw_talker")
+  {
+    raw_msg_ = rmw_get_zero_initialized_raw_message();
+    auto allocator = rcutils_get_default_allocator();
+    rmw_initialize_raw_message(
+      &raw_msg_,
+      0u,
```
sloretz: Maybe use a variable to indicate what the number is?

```c
const unsigned int capacity = 0u;
rwm_initialize_raw_message(
   &raw_msgs_,
   capacity,
   &allocator);
```

demo_nodes_cpp/src/topics/talker_raw.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    // Create a function for when messages are to be sent.
+    auto publish_message =
+      [this]() -> void
+      {
+        auto string_msg = std::make_shared<std_msgs::msg::String>();
+        string_msg->data = "Hello World:" + std::to_string(count_++);
+        auto string_ts =
+          rosidl_typesupport_cpp::get_message_type_support_handle<std_msgs::msg::String>();
+        rmw_raw_message_resize(&raw_msg_, 8u + static_cast<unsigned int>(string_msg->data.size()));
```
sloretz: What's the extra 8 bytes for?

sloretz: Is calling `rmw_raw_message_resize()` here necessary? It looks like `rmw_fastrtps`'s implementation  of `rmw_serialize` will resize if necessary.

https://github.com/ros2/rmw_fastrtps/pull/186/files#diff-b260f922d969aa4536762ca6b4dfe86dR51

demo_nodes_cpp/src/topics/talker_raw.cpp
```diff
@@ -0,0 +1,137 @@
...
+        auto ret = rmw_serialize(string_msg.get(), string_ts, &raw_msg_);
+        if (ret != RMW_RET_OK) {
+          fprintf(stderr, "failed to deserialize raw message\n");
```
sloretz: nit: deserialize -> serialize

demo_nodes_cpp/src/topics/talker_raw.cpp
```diff
@@ -0,0 +1,137 @@
...
+        printf("ROS message:\n");
+        printf("%s\n", string_msg->data.c_str());
+        printf("Raw message:\n");
+        for (unsigned int i = 0; i < raw_msg_.buffer_length; ++i) {
+          printf("%02x ", raw_msg_.buffer[i]);
+        }
+        printf("\n");
+
+        // Put the message into a queue to be processed by the middleware.
+        // This call is non-blocking.
```
sloretz: Is it? ros2/ros2#255 . Maybe say `assume this call is non-blocking`?

demo_nodes_cpp/src/topics/allocator_tutorial.cpp
```diff
@@ -169,7 +169,7 @@ int main(int argc, char ** argv)
     rclcpp::message_memory_strategy::MessageMemoryStrategy<
       std_msgs::msg::UInt32, MyAllocator<>>>(alloc);
   auto subscriber = node->create_subscription<std_msgs::msg::UInt32>(
-    "allocator_tutorial", 10, callback, nullptr, false, msg_mem_strat, alloc);
+    "allocator_tutorial", callback, 10, nullptr, false, msg_mem_strat, alloc);
```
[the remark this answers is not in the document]
Karsten1987: with the changes to the subscription_traits I had to swap the two arguments in position. See here: https://github.com/ros2/rclcpp/pull/388#discussion_r173047624

demo_nodes_cpp/src/topics/talker_raw.cpp
Karsten1987: yeah. Maybe I should make the demo a bit specific. The idea here is that with preallocating the memory, no dynamic memory allocation is needed anymore and the complete pipeline down to the actual publishing is free of any memory allocation.
I'll change the code to make that more visible.

demo_nodes_cpp/src/topics/listener_raw.cpp
Karsten1987: we're currently using `setvbuf` in other demo examples as well. So i'll keep it consistent with them.

demo_nodes_cpp/src/topics/listener_raw.cpp
Karsten1987: I decided to go with `rmw_*` everywhere as I also call the `rmw_serialize` functions directly.

Karsten1987:
So far I've only seen an extra `00` at the end, which I interpret as the null termination done by CDR. I can't really reproduce the 2x `00`. Does this happen with all DDS implementations?

sloretz:
@Karsten1987 I've only tried it with fast-rtps. The number of extra null bytes depends on the length of the string. At 9 there is a two byte difference. At 10 just a 1 bytes difference.

Talker
```
Hello World:9
Raw message:
00 01 00 00 0e 00 00 00 48 65 6c 6c 6f 20 57 6f 72 6c 64 3a 39 00 
ROS message:
Hello World:10
Raw message:
00 01 00 00 0f 00 00 00 48 65 6c 6c 6f 20 57 6f 72 6c 64 3a 31 30 00
```

Listener
```
had to resize to 24
I heard raw data of length: 24
00 01 00 00 0e 00 00 00 48 65 6c 6c 6f 20 57 6f 72 6c 64 3a 39 00 00 00 
Raw data after deserialization: Hello World:9
had to resize to 24
I heard raw data of length: 24
00 01 00 00 0f 00 00 00 48 65 6c 6c 6f 20 57 6f 72 6c 64 3a 31 30 00 00 
Raw data after deserialization: Hello World:10
```

Karsten1987:
new CI:

linux: [Build Status](https://ci.ros2.org/job/ci_linux/4485/)
osx: [Build Status](https://ci.ros2.org/job/ci_osx/3674/)
win: [Build Status](https://ci.ros2.org/job/ci_windows/4560/)

Karsten1987:
@sloretz do you mind giving this another look? I'd really appreciate getting this set of PRs in before bouncy code freeze. Thanks :)

wjwwood:
I put this back into "in progress" to keep the review column tidy, please put it back in "review" when the feedback is addressed.

demo_nodes_cpp/src/topics/listener_raw.cpp
```diff
@@ -0,0 +1,115 @@
...
+        // In order to deserialize the message we have to manually create a ROS2
+        // message in which we want to convert the raw data.
+        auto string_msg = std::make_shared<std_msgs::msg::String>();
+        auto string_ts =
+          rosidl_typesupport_cpp::get_message_type_support_handle<std_msgs::msg::String>();
+        // The rmw_deserialize fucntion takes the raw data and a corresponding typesupport
+        // which is responsible on how to convert this data into a ROS2 message.
+        auto ret = rmw_deserialize(msg.get(), string_ts, string_msg.get());
+        if (ret != RMW_RET_OK) {
+          fprintf(stderr, "failed to deserialize raw message\n");
```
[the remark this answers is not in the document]
Karsten1987: absolutely right. I added a return in https://github.com/ros2/demos/pull/185/commits/cdf33e5f21f55509ffa7bdbc1cf33ed9b7046ea3 to keep the loop going in case it was just a hiccup.

demo_nodes_cpp/src/topics/talker_raw.cpp
```diff
@@ -0,0 +1,159 @@
...
+        auto message_payload_length = static_cast<unsigned int>(string_msg->data.size());
+        rmw_raw_message_resize(&raw_msg_, message_header_length + message_payload_length);
+
+        auto string_ts =
+          rosidl_typesupport_cpp::get_message_type_support_handle<std_msgs::msg::String>();
+        // Given the correct typesupport, we can convert our ROS2 message into
+        // its binary representation (raw_msg)
+        auto ret = rmw_serialize(string_msg.get(), string_ts, &raw_msg_);
+        if (ret != RMW_RET_OK) {
+          fprintf(stderr, "failed to serialize raw message\n");
```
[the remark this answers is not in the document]
Karsten1987: see https://github.com/ros2/demos/pull/185/commits/cdf33e5f21f55509ffa7bdbc1cf33ed9b7046ea3

wjwwood:
I've re-reviewed all the pull requests and so here's some CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4670/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1557/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3861/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4742/)

Karsten1987:
linux: [Build Status](https://ci.ros2.org/job/ci_linux/4671/)
osx: [Build Status](https://ci.ros2.org/job/ci_osx/3862/)
aarch: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/1558/)
win: [Build Status](https://ci.ros2.org/job/ci_windows/4749/)


I further opened a PR against `rmw_opensplice_cpp` which implements the missing functions and sets the error message accordingly when calling them.

wjwwood:
Rerun of Windows CI after https://github.com/ros2/rmw_connext/pull/259/commits/8eaeb9e7e8899a4797810fbcf8919b1c9d5867a0: [Build Status](https://ci.ros2.org/job/ci_windows/4765/)

wjwwood:
Rerun of Windows CI after ros2/rmw_connext@f778624: [Build Status](https://ci.ros2.org/job/ci_windows/4771/)

wjwwood:
Rerun of Windows CI after ros2/rmw_connext@27cd7636aaa89d76002cd11cffa8bb54082a72d8: [Build Status](https://ci.ros2.org/job/ci_windows/4772/)

Karsten1987:
So I merged this! 
Thanks @wjwwood for finishing up the last hiccup on Windows! Appreciated!

PR #186, opened 2017-10-24
updating to use new time source logic

tfoote:
Connects to ros2/rcl#164

### DIFF
--- a/dummy_robot/dummy_map_server/src/dummy_map_server.cpp
+++ b/dummy_robot/dummy_map_server/src/dummy_map_server.cpp
@@ -19,7 +19,7 @@
 #include "rclcpp/clock.hpp"
 #include "rclcpp/rclcpp.hpp"
 #include "rcl/rcl.h"
-
+#include "rclcpp/time_source.hpp"
 
 #include "nav_msgs/msg/occupancy_grid.hpp"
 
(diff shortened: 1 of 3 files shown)


dummy_robot/dummy_map_server/src/dummy_map_server.cpp
```diff
@@ -64,7 +66,8 @@ int main(int argc, char * argv[])
     msg->data[(++center) % (msg->info.width * msg->info.height)] = 100;
     msg->data[(++rhs) % (msg->info.width * msg->info.height)] = 0;
 
-    msg->header.stamp = rclcpp::Time::now();
+    rclcpp::Clock ros_clock(RCL_ROS_TIME);
+    msg->header.stamp = ros_clock.now();
```
wjwwood: Unlike in https://github.com/ros2/geometry2/pull/49, this code doesn't seem to include the node or time source in here anywhere, is that intentional?

dummy_robot/dummy_map_server/src/dummy_map_server.cpp
```diff
@@ -64,7 +66,10 @@ int main(int argc, char * argv[])
     msg->data[(++center) % (msg->info.width * msg->info.height)] = 100;
     msg->data[(++rhs) % (msg->info.width * msg->info.height)] = 0;
 
-    msg->header.stamp = rclcpp::Time::now();
+    rclcpp::TimeSource ts(node);
+    rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
+    ts.attachClock(clock);
+    msg->header.stamp = clock->now();
```
mikaelarguedas: @tfoote Can we move all the instantiation outside of the while loop and keep only the call to `clock->now()` here? (Same below)

tfoote: Yes we should do that.

PR #189, opened 2017-11-14
 Strip formatting from console output for output matching

dhood:
This is preparation for swapping demos to logging macros. I don't mind if we merge this now or later but I want to kick off the review process in case there are any hiccups.

29e3716eb1d6c363f70bbfbc9994a7ca91ae2c56 is included to show that the change does what's expected, it's not being proposed for merge and will be removed before merge.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3511/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/731/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2847/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3574/)

### DIFF
--- a/composition/src/listener_component.cpp
+++ b/composition/src/listener_component.cpp
@@ -17,6 +17,7 @@
 #include <iostream>
 #include <memory>
 
+#include "rcutils/logging_macros.h"
 #include "rclcpp/rclcpp.hpp"
 #include "std_msgs/msg/string.hpp"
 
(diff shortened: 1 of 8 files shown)


dhood:
merging now that the functionality is exercised by https://github.com/ros2/demos/pull/191

dirk-thomas:
These `setup()` function aren't used when the generated Python file is being invoked directly (using `__main__`). Removed as part of ros2/demos#225.

PR #190, opened 2017-11-15
 Replace print() in python demos with logger calls

dhood:
Connects to ros2/rclpy#148
Requires ros2/rclpy#148

### DIFF
--- a/demo_nodes_py/demo_nodes_py/topics/listener_qos.py
+++ b/demo_nodes_py/demo_nodes_py/topics/listener_qos.py
@@ -20,6 +20,8 @@ from rclpy.qos import qos_profile_default, qos_profile_sensor_data
 
 from std_msgs.msg import String
 
+logger = rclpy.logging.get_named_logger('listener_qos')
+
 
 class ListenerQos(rclpy.Node):
 
@@ -29,7 +31,7 @@ class ListenerQos(rclpy.Node):
             String, 'chatter', self.chatter_callback, qos_profile=qos_profile)
 
     def chatter_callback(self, msg):
-        print('I heard: [%s]' % msg.data)
+        self.get_logger().info('I heard: [%s]' % msg.data)
 
 
 def main(argv=sys.argv[1:]):
@@ -46,10 +48,10 @@ def main(argv=sys.argv[1:]):
 
     if args.reliable:
         custom_qos_profile = qos_profile_default
-        print('Reliable listener')
+        logger.info('Reliable listener')
     else:
         custom_qos_profile = qos_profile_sensor_data
-        print('Best effort listener')
+        logger.info('Best effort listener')
 
     node = ListenerQos(custom_qos_profile)
 
--- a/demo_nodes_py/demo_nodes_py/topics/listener_qos.py
+++ b/demo_nodes_py/demo_nodes_py/topics/listener_qos.py
@@ -17,16 +17,19 @@ import sys
 
 import rclpy
 from rclpy.qos import qos_profile_default, qos_profile_sensor_data
+from rclpy.qos import QoSReliabilityPolicy
 
 from std_msgs.msg import String
-
-logger = rclpy.logging.get_named_logger('listener_qos')
 
 
 class ListenerQos(rclpy.Node):
 
     def __init__(self, qos_profile):
         super().__init__('listener_qos')
+        if qos_profile.reliability is QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE:
+            self.get_logger().info('Reliable listener')
+        else:
+            self.get_logger().info('Best effort listener')
         self.sub = self.create_subscription(
             String, 'chatter', self.chatter_callback, qos_profile=qos_profile)
 
@@ -48,10 +51,8 @@ def main(argv=sys.argv[1:]):
 
     if args.reliable:
         custom_qos_profile = qos_profile_default
-        logger.info('Reliable listener')
     else:
         custom_qos_profile = qos_profile_sensor_data
-        logger.info('Best effort listener')
 
     node = ListenerQos(custom_qos_profile)
 
--- a/demo_nodes_py/demo_nodes_py/topics/talker_qos.py
+++ b/demo_nodes_py/demo_nodes_py/topics/talker_qos.py
@@ -28,9 +28,9 @@ class TalkerQos(rclpy.Node):
         super().__init__('talker_qos')
         self.i = 0
         if qos_profile.reliability is QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE:
-            self.get_logger().info('Reliable listener')
+            self.get_logger().info('Reliable talker')
         else:
-            self.get_logger().info('Best effort listener')
+            self.get_logger().info('Best effort talker')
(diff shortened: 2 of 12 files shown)


demo_nodes_py/demo_nodes_py/topics/listener_qos.py
```diff
@@ -25,11 +26,15 @@ class ListenerQos(rclpy.Node):
 
     def __init__(self, qos_profile):
         super().__init__('listener_qos')
+        if qos_profile.reliability is QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE:
+            self.get_logger().info('Reliable listener')
```
mikaelarguedas: It's not obvious to me that this is actually performing a logging action compared to the API suggested originally in https://github.com/ros2/rclpy/issues/103, maybe just calling the function `loginfo` or a `log` function taking a severity keyword would be clearer.
I know the recent discussion was focused on the c++ API but how do we expect the API to look like for the users in python?

dhood: the `.log(,INFO)` option is also available on this logger, following https://github.com/ros2/rclpy/issues/103#issuecomment-324148790 (which came from a team discussion), `.info()` is offered for convenience. Both options mirror what is offered by python logging, we seem to prefer the `.info()`: https://github.com/ros2/ros2cli/blob/9fce6a08381bb67cf17e7160b0b2c6556ffec1e6/ros2cli/ros2cli/entry_points.py#L87

mikaelarguedas: Ok so the goal is to use an API as close as python native logging as possible rather than something closer to ROS 1. We do provide both but will use this one in our examples. Fair enough.
With https://github.com/ros2/rclpy/pull/148#issuecomment-344781872 it now make sense why we don't something like `self.mylogger.loginfo()`

demo_nodes_py/demo_nodes_py/topics/talker_qos.py
```diff
@@ -26,15 +27,20 @@ class TalkerQos(rclpy.Node):
     def __init__(self, qos_profile):
         super().__init__('talker_qos')
         self.i = 0
+        if qos_profile.reliability is QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE:
+            self.get_logger().info('Reliable listener')
+        else:
+            self.get_logger().info('Best effort listener')
```
sloretz: `listener` -> `talker`

dhood:
> It's not caused by this PR, but it seems like the logged messages are really long.

https://github.com/ros2/rcutils/pull/72 shortens the default console output formatting

demo_nodes_py/demo_nodes_py/topics/talker_qos.py
dhood: thanks: bbdcaa3d5d65059ef2f1cf8efcbce6df27a49006

demo_nodes_py/demo_nodes_py/topics/listener_qos.py
dhood: Following up on this, if I understand correctly your initial concern was referring to how we would use the generic logging functions (e.g. `rclpy.logging.loginfo()`) as opposed to log calls being made on specific logger objects (which weren't around in ROS 1).

On the use of `mylogger.info()` as opposed to `mylogger.log(, severity=INFO)`, as you noted I've chosen to mirror the typical usage of native python logging. https://docs.python.org/3/howto/logging-cookbook.html, for example, uses `mylogger.info()` in all cases except where the severity is chosen programatically, for which case the `mylogger.log(, severity=INFO)` form is available

mikaelarguedas:
Is this ready for review?

dhood:
yes, I moved the connected tickets across and asked for its review in https://github.com/ros2/rclpy/pull/148 but the label didn't update here

lifecycle/src/lifecycle_service_client_py.py
```diff
@@ -75,24 +79,24 @@ def get_available_transitions(lifecycle_node):
...
+            % (transition.start_state.id, transition.start_state.label))
+        node.get_logger().info(
+            '\tGoal  id: %u\tlabel: %s'
+            % (transition.goal_state.id, transition.goal_state.label))
 
 
 def main(service_type, lifecycle_node, change_state_args='', args=None):
     rclpy.init(args=args)
 
-    if not rclpy.ok():
```
mikaelarguedas: This is not necessary anymore because init would raise on failure ?

topic_monitor/topic_monitor/launch_files/launch_reliability_demo.py
```diff
@@ -25,6 +26,10 @@ def main():
     package = 'topic_monitor'
     executable = get_executable_path(package_name=package, executable_name='data_publisher')
 
+    # Strip the logger name from the message format in favor of the shorter executable name
+    os.environ['RCUTILS_CONSOLE_OUTPUT_FORMAT'] = '[{severity}] {message}'
```
mikaelarguedas: Does this modified environment need to be passed to the launched processes ? (maybe it's not necessary as the same environment is used for the launchfile and each launched process). Otherwise there's an example of custom environment passed to launched processes here:
https://github.com/ros2/system_tests/blob/a4e81e348f7d2b25076f2b11dfde0f073e3850de/test_communication/test/test_publisher_subscriber.py.in#L28-L32

lifecycle/src/lifecycle_service_client_py.py
dhood: yep, thanks for making it explicit :smile:

topic_monitor/topic_monitor/launch_files/launch_reliability_demo.py
dhood: yeah any children processes will inherit this environment automatically

mikaelarguedas: :+1:

topic_monitor/topic_monitor/scripts/topic_monitor.py
```diff
@@ -24,6 +25,7 @@
 from std_msgs.msg import Float32, Header
 
 QOS_DEPTH = 10
+logger = rclpy.logging.get_named_logger('topic_monitor')
```
dirk-thomas: This line fails for me with:

```
AttributeError: module 'rclpy' has no attribute 'logging'
```

dhood: sounds like another side effect of https://github.com/ros2/rclpy/pull/147, will fix

dirk-thomas: Well, the referenced PR was merged way before this one. So this change hasn't worked when is was merged.

dhood: not trying to play the blame game here, just to understand and document what went wrong. The reason why that PR is relevant is because it broke many demos without us realising (and this PR was tested before that one was opened), and we intend to [investigate how to undo the side effects](https://github.com/ros2/rclpy/issues/156). therefore knowing that this is one of them is important

dhood: https://github.com/ros2/demos/pull/210

PR #191, opened 2017-11-16
Use logging macros [logger name instead of object]

dhood:
This currently has a few files converted: 4 commits for 4 different "types" of conversion.

Please give feedback on this and then I'll do the rest of them matching the conventions.

Remember that the macros are taking a name instead of a logger object at the moment: `node->get_name()` will one day be replaced by `node->get_logger()`.

### DIFF
--- a/demo_nodes_cpp/src/topics/listener.cpp
+++ b/demo_nodes_cpp/src/topics/listener.cpp
@@ -21,6 +21,8 @@
 
 #include "std_msgs/msg/string.hpp"
 
+const char * g_logger_name = nullptr;
+
 void print_usage()
 {
   printf("Usage for listener app:\n");
@@ -32,13 +34,14 @@ void print_usage()
 
 void chatterCallback(const std_msgs::msg::String::SharedPtr msg)
 {
-  std::cout << "I heard: [" << msg->data << "]" << std::endl;
+  RCLCPP_INFO(g_logger_name, "I heard: [%s]", msg->data.c_str());
 }
 
 int main(int argc, char * argv[])
 {
   rclcpp::init(argc, argv);
   auto node = rclcpp::Node::make_shared("listener");
+  g_logger_name = node->get_name();
 
   if (rcutils_cli_option_exist(argv, argv + argc, "-h")) {
     print_usage();
--- a/demo_nodes_cpp/src/topics/listener.cpp
+++ b/demo_nodes_cpp/src/topics/listener.cpp
@@ -21,8 +21,6 @@
 
 #include "std_msgs/msg/string.hpp"
 
-const char * g_logger_name = nullptr;
-
 void print_usage()
 {
   printf("Usage for listener app:\n");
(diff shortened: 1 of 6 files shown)


demo_nodes_cpp/src/topics/listener.cpp
```diff
@@ -32,13 +34,14 @@ void print_usage()
 
 void chatterCallback(const std_msgs::msg::String::SharedPtr msg)
 {
-  std::cout << "I heard: [" << msg->data << "]" << std::endl;
+  RCLCPP_INFO(g_logger_name, "I heard: [%s]", msg->data.c_str());
```
dhood: I'm not really a fan of this but it's the most direct approach for demos with one node. Other options here are to change the callback signature and use std::bind (it will be a bit clunky because of https://github.com/ros2/rclcpp/issues/273, so I avoided it for our "minimalist" demo) or add a standard callback signature to rclcpp that includes the logger, but if everyone is using composition then that signature "shouldn't" be needed.

Speaking of which, it's probably time to convert these demos to at least inherit from node, which will remove this less-than-ideal logging practice. I'll do that.

dhood: I opened https://github.com/ros2/demos/pull/192 and added bd071b53f5380db71501d82d2957523bf2b89540 as a less-controversial way to pass "node loggers" to free function

dhood:
To make reviewing easier, rather than change all of the files at once, I've done a sample of files, and then I'll propagate the conventions to the rest of them. Please see the individual commits for 4 different "types" of conversion.

dhood:
Another point to take note of: in the past we've been against semicolons on macros.

Since we will likely require semicolons at the end of macros in the future (e.g. to limit the scope of locally defined variables as is done in [other rcutils macros](https://github.com/ros2/rcutils/pull/63/files#diff-1b06d4a1ccca0f0dff66d961923143a1R129) and in [rosconsole logging macros](https://github.com/ros/ros_comm/blob/4383f8fad9550836137077ed1a7120e5d3e745de/tools/rosconsole/include/ros/console.h#L553)), I have not removed them from these logging macro calls.

dhood:
Never mind, I don't want to hold this up bringing in more discussion points than necessary (plus, limiting the scope does not always imply requiring a semicolon). I've removed the semicolons in https://github.com/ros2/demos/pull/191/commits/fb69a377a7f90a877ea49eeb5eeb7dec1c6a4254 so the logging macro calls are semi-colon-less as in other parts of the code base.

Please review, @ros2/team

dhood:
> I guess we want this to be merged before the alternative approach suggested in #192 is merged?

It's the other way around (for the demos that need to be swapped to inheritance; the demos in this PR don't need to be changed so there is no 'blocker').

Thanks for the reviews on the subset of files: after CI with https://github.com/ros2/demos/pull/189 I will merge this PR so we don't have to re-review these parts, but there will be a followup PR including the changes to the other demos.

dhood:
CI including https://github.com/ros2/demos/pull/189:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3574/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/771/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2913/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3667/)

PR #192, opened 2017-11-16
C++ demo nodes using inheritance

dhood:
I want to swap to using logging macros in https://github.com/ros2/demos/pull/191.
Having free functions was complicating the process, but these demos should be using Node inheritance anyway.

haven't done the service demos yet


dhood:
Ready for review. The point of this PR is to get rid of free functions so that the demo nodes are set up for using logging macros how we expect them to be used (accessing the node's logger). There are still some nodes that don't inherit from Node in this repo but not in the demo_node_cpp package, with the exception of parameter nodes that are blocked by https://github.com/ros2/rclcpp/issues/409.

I left the sync client without node inheritance because there isn't really a distinction between sync and async when [using a response callback function](https://github.com/ros2/demos/pull/192/files#diff-6ecbd65ded64c0bc1c96360a682c40f7R77) (which is required if spin has been called on the node by the main). The `spin_until_future_complete` usage is covered by https://github.com/ros2/examples/blob/1907bac059ba4962bbeb0a2cf8f3c5f84cdbab95/rclcpp/minimal_client/main.cpp#L39, if we want to remove the sync client from this package.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3622/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/804/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2952/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3711/)

dhood:
@ros2/team waiting for review. If you notice inconsistency in the usage of `rclcpp::Subscription` vs `rclcpp::service::Service` it's related to https://github.com/ros2/demos/pull/199 and can be addressed after https://github.com/ros2/rclcpp/pull/410 is merged

demo_nodes_cpp/src/services/add_two_ints_client_async.cpp
```diff
@@ -27,51 +28,85 @@ using namespace std::chrono_literals;
...
+  explicit ClientNode(const std::string & service_name)
+  : Node("add_two_ints_client")
+  {
+    client_ = create_client<example_interfaces::srv::AddTwoInts>(service_name);
+    auto on_timer =
+      [this]() -> void {
+        timer_->cancel();  // We only want to make a single request.
+        this->make_request();
+      };
+    timer_ = create_wall_timer(0s, on_timer);  // This will trigger once spin is called on the node.
```
dhood: @Karsten1987 and I spoke about this and it should be fine to just call `make_request` straight away here (should rename the method to `start_async_request` or something) since both approaches just queue something for once the node gets spun on.. I'll give it a shot

dhood: simplified in a010a4dc9021fc658acacbde10632d2840ec3629 @mikaelarguedas fyi (you also mentioned this was unnecessarily complicated FWIU)

demo_nodes_cpp/src/services/add_two_ints_client_async.cpp
```diff
@@ -27,51 +28,76 @@ using namespace std::chrono_literals;
...
+      };
+    auto future_result = client_->async_send_request(request, response_received_callback);
   }
 
-  auto request = std::make_shared<example_interfaces::srv::AddTwoInts::Request>();
-  request->a = 2;
-  request->b = 3;
+private:
+  rclcpp::client::Client<example_interfaces::srv::AddTwoInts>::SharedPtr client_;
+  rclcpp::timer::TimerBase::SharedPtr timer_;
```
Karsten1987: we don't need that `timer_` anymore, do we?

dhood: absolutely right, thanks! b9e57944dee4f88e9030fb1d1cb4233f7e1efd4e

dhood:
>  If you notice inconsistency in the usage of rclcpp::Subscription vs rclcpp::service::Service it's related to #199 and can be addressed after ros2/rclcpp#410 is merged

rebased since those PRs have been merged and removed any extra namespaces so you shouldn't see e.g. `rclcpp::service::Service` in this PR anymore

dhood:
CI (planning to merge when comes back green):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3693/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/860/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3028/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3789/)

PR #196, opened 2017-11-27
Fix import statement and usage for rclpy.node.Node

mikaelarguedas:
node is not imported in `__init__.py` anymore (https://github.com/ros2/rclpy/pull/147)

Without this change the demos raise:
`AttributeError: module 'rclpy' has not attribute 'Node'`

Not sure if https://github.com/ros2/rclpy/pull/147 is a long term fix as this mean that anyone inheriting from Node will need to import it explicitly.

### DIFF
--- a/demo_nodes_py/demo_nodes_py/services/add_two_ints_server.py
+++ b/demo_nodes_py/demo_nodes_py/services/add_two_ints_server.py
@@ -15,9 +15,10 @@
 from example_interfaces.srv import AddTwoInts
 
 import rclpy
+from rclpy.node import Node
 
 
-class AddTwoIntsServer(rclpy.Node):
+class AddTwoIntsServer(Node):
 
     def __init__(self):
         super().__init__('add_two_ints_server')
--- a/demo_nodes_py/demo_nodes_py/topics/listener.py
+++ b/demo_nodes_py/demo_nodes_py/topics/listener.py
@@ -15,11 +15,12 @@
 import sys
 
 import rclpy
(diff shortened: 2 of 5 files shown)


wjwwood:
> Sure, do you have a suggestion?

"Fix import statement and usage for rclpy.node.Node"?

----

This is really unfortunate because it adds another line to all of the examples unnecessarily (imo). It is also asymmetric with what we want to do with the C++ API, i.e. we want `rclcpp::Node` rather than `rclcpp::node::Node`. I don't understand fully why @dirk-thomas needed to do what he did in https://github.com/ros2/rclpy/pull/147#discussion_r151004022, but I think it's a step backwards for the usability of the `rclpy` API.

mikaelarguedas:
> "Fix import statement and usage for rclpy.node.Node"?

Sounds good to me 

---

> This is really unfortunate because it adds another line to all of the examples unnecessarily (imo). It is also asymmetric with what we want to do with the C++ API, i.e. we want rclcpp::Node rather than rclcpp::node::Node. I don't understand fully why @dirk-thomas needed to do what he did in ros2/rclpy#147 (comment), but I think it's a step backwards for the usability of the rclpy API.

Yeah I agree, I'm proposing this fix mostly to get the demos back to a working state but don't think users should have to import this explicitly in the long-term.
I'd like to merge this before next week but can open a follow-up issue to revisit this before the release. How does that sound ?

wjwwood:
+1 for getting things working again, but we should try to address this before the release I think.

PR #197, opened 2017-11-28
Remove 'node' namespace from rclcpp::Node

dhood:
Based off https://github.com/ros2/demos/pull/196#issuecomment-347325825:
> what we want to do with the C++ API, i.e. we want rclcpp::Node rather than rclcpp::node::Node.

---

The style I chose is to match https://github.com/ros2/demos/blob/5ba0e66d31374815fb6f0bfb279b4d982459d940/intra_process_demo/src/cyclic_pipeline/cyclic_pipeline.cpp#L30

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3618/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/801/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2948/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3705/)

### DIFF
--- a/demo_nodes_cpp/src/timers/one_off_timer.cpp
+++ b/demo_nodes_cpp/src/timers/one_off_timer.cpp
@@ -24,7 +24,7 @@ class OneOffTimerNode : public rclcpp::Node
 {
 public:
   OneOffTimerNode()
-  : rclcpp::Node("one_off_timer"), count(0)
+  : Node("one_off_timer"), count(0)
   {
     periodic_timer = this->create_wall_timer(
       2s,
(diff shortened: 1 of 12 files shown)


demo_nodes_cpp/src/timers/one_off_timer.cpp
```diff
@@ -24,7 +24,7 @@ class OneOffTimerNode : public rclcpp::Node
 {
 public:
   OneOffTimerNode()
-  : rclcpp::Node("one_off_timer"), count(0)
+  : Node("one_off_timer"), count(0)
```
wjwwood: This is fine (correct) as a shortcut, but the existing code is also fine and a bit more explicit. I don't think you need to change this for this pr, but I wouldn't necessarily say we need to go out and do a similar change everywhere. Both are correct, and I think it's fine to leave it up to the author to decide which to use in the situation.

dhood: OK, no worries, I'll go easy on these ones :smile: thanks

PR #198, opened 2017-11-29
member of rosidl_interfaces_packages group

mikaelarguedas:
connects to ros2/common_interfaces#45

### DIFF
--- a/composition/package.xml
+++ b/composition/package.xml
@@ -32,7 +32,7 @@
   <test_depend>launch_testing</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
 
-  <member_of_group>rosidl_interfaces_packages</member_of_group>
+  <member_of_group>rosidl_interface_packages</member_of_group>
 
   <export>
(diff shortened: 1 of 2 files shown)


composition/package.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
-<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
```
dhood: do we expect this to exist yet? (my linter's unhappy)

mikaelarguedas: no it doesnt exist yet. It will when we ratify the rep, in the meantime xmllint will complain but it won't impact any of our tools (https://github.com/ros2/rmw_implementation/pull/32#discussion_r148157220)

PR #199, opened 2017-11-30
Use escalated namespaces for more entities

dhood:
connects to ros2/rclcpp#410
requires ros2/rclcpp#410

Demos were using `rclcpp::Publisher` but `rclcpp::service::Service` (because it hadn't been escalated)


dhood:
CI is in https://github.com/ros2/rclcpp/pull/410

PR #205, opened 2017-12-05
Use logging macros in more demos

dhood:
Followup of https://github.com/ros2/demos/pull/191 where a few were changed as prototypes

CI linux arm [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/881/) (still checking that the tests pass)

Needed to update expected output and add force flush for some nodes.

### DIFF
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -101,7 +101,7 @@ int main(int argc, char * argv[])
 
   // Initialize a ROS 2 node to publish images read from the OpenCV interface to the camera.
   auto node = rclcpp::Node::make_shared("cam2image");
-  auto node_logger = node->get_logger();
+  rclcpp::Logger node_logger = node->get_logger();
 
   // Set the parameters of the quality of service profile. Initialize as the default profile
   // and set the QoS parameters specified on the command line.
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -101,6 +101,7 @@ int main(int argc, char * argv[])
 
   // Initialize a ROS 2 node to publish images read from the OpenCV interface to the camera.
   auto node = rclcpp::Node::make_shared("cam2image");
+  auto node_logger = node->get_logger();
 
   // Set the parameters of the quality of service profile. Initialize as the default profile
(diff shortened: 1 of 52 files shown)


dhood:
I'll look into why test output doesn't match correctly (probably missed some force flush of stdout) if anyone wants to review the logger usage in the meantime

image_tools/src/cam2image.cpp
```diff
@@ -101,7 +101,7 @@ int main(int argc, char * argv[])
 
   // Initialize a ROS 2 node to publish images read from the OpenCV interface to the camera.
   auto node = rclcpp::Node::make_shared("cam2image");
-  auto node_logger = node->get_logger();
+  rclcpp::Logger node_logger = node->get_logger();
```
wjwwood: Out of curiosity, why do you think that?

dhood: in http://ci.ros2.org/job/ci_windows/3833/ it failed at the decltype check in the macro. it's just a guess though :)

wjwwood: I see, that might be, I didn't consider it was interacting with decltype, but even then it _should_ work :). Hopefully you'll find the issue.

dhood: thanks to your pointers I opened https://github.com/ros2/rclcpp/pull/423 and the usage in the lambda below can pass the type check

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3738/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/890/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3080/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3833/)

windows requires https://github.com/ros2/rclcpp/pull/423. CI for windows including https://github.com/ros2/rclcpp/pull/423 [Build Status](http://ci.ros2.org/job/ci_windows/3838/)

dhood:
more ci including https://github.com/ros2/rclcpp/pull/423
linux [Build Status](http://ci.ros2.org/job/ci_linux/3744/)
osx [Build Status](http://ci.ros2.org/job/ci_osx/3084/)

PR #206, opened 2017-12-05
use sstream rather than iostream

mikaelarguedas:
noticed this when going through the PR.
Feel free to merge or discard as this is not required to build or pass the linters


dhood:
hmmm ostream isn't necessary to bring in endl? that was why I hadn't swap
thanks for the initiative on this either way

mikaelarguedas:
I didn't look into it in detail and I think is not worthe spending time on. Feel free to close it, my approval of #205 still stands

dhood:
I don't think we want the endl anyway so I removed the usage

PR #207, opened 2017-12-05
add missing newline

mikaelarguedas:
this should fix the failing test


dirk-thomas:
Can you please explain why that should be necessary? When running the native talker I don't see any empty lines. Also if the test hasn't changed it shouldn't start to fail.

mikaelarguedas:
This PR and #206 are targeting the `more_logging` branch, not master

dhood:
thanks! I'd left the newline by accident: I removed it in https://github.com/ros2/demos/pull/205/commits/47752ea6a6a664d98ced7148275509a7924cd4cf#diff-b798911dea84b9222151472238cdaf13L58 which ought to fix the test matching without needing to update the expected output. thanks for looking into it though

Issue #213, opened 2017-12-07
Trying to get the state of lifecycle listener hangs forever


clalancette:
## Bug report

I'm running the tutorial at https://github.com/ros2/ros2/wiki/managed-nodes#run-the-demo on Linux ARM64 with the Fast-RTPS only fat archive.  If I run the `lifecycle_talker`, then try `get_state` with `lifecycle_service_client_py.py`, I get the correct state.  If I then run `lifecycle_listener`, and try to get the state using `lifecycle_service_client_py.py`, it hangs forever.  When I eventually hit Ctrl-C, it throws a traceback:

```
^CTraceback (most recent call last):
  File "/home/ubuntu/ros2-linux/lib/lifecycle/lifecycle_service_client_py.py", line 125, in <module>
    main(args.service, args.node, args.change_state_args)
  File "/home/ubuntu/ros2-linux/lib/lifecycle/lifecycle_service_client_py.py", line 104, in main
    get_state(lifecycle_node)
  File "/home/ubuntu/ros2-linux/lib/lifecycle/lifecycle_service_client_py.py", line 58, in get_state
    % (lifecycle_node, cli.response.current_state.label, cli.response.current_state.id))
AttributeError: 'NoneType' object has no attribute 'current_state'
Exception ignored in: <bound method Node.__del__ of <rclpy.node.Node object at 0x7fa6622860>>
Traceback (most recent call last):
  File "/home/ubuntu/ros2-linux/lib/python3.5/site-packages/rclpy/node.py", line 300, in __del__
    self.destroy_node()
  File "/home/ubuntu/ros2-linux/lib/python3.5/site-packages/rclpy/node.py", line 275, in destroy_node
    _rclpy.rclpy_destroy_node_entity(cli.client_handle, self.handle)
RuntimeError: Failed to fini 'rcl_client_t': rcl node is invalid, rcl instance id does not match, at /home/rosbuild/ci_scripts/ws/src/ros2/rcl/rcl/src/rcl/node.c:382
```

**Required Info:**

- Operating System:
  - Ubuntu 16.04 ARM64
- Installation type:
  - Fast-RTPS-only fat archive
- Version or commit hash:
  - c44e525cb9c3c13b3f5dd04284e28f0d76b36b55 of demos repo
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp/rclpy

#### Steps to reproduce issue

Terminal 1:
```
ros2 run lifecycle lifecycle_talker
```

[...]

PR #214, opened 2017-12-07
add wait_for_service

Karsten1987:
Fixes https://github.com/ros2/demos/issues/213

Call wait_for_service prior to calling it to make sure the lifecycle service is available. In the case of calling a node wo/ this service available, the request times out after 5.0 seconds and prints an error message.

```
$ ros2 run lifecycle lifecycle_service_client_py.py -- get_state lc_listener
Unable to call service lc_listener/get_state
```

@clalancette can you verify this please?

### DIFF
--- a/lifecycle/src/lifecycle_service_client_py.py
+++ b/lifecycle/src/lifecycle_service_client_py.py
@@ -27,7 +27,8 @@ def change_state(lifecycle_node, change_state_args=''):
     service_name = lifecycle_node + '/change_state'
     cli = node.create_client(ChangeState, service_name)
     if not cli.wait_for_service(timeout_sec=5.0):
-        print('Unable to call service', service_name)
+        node.get_logger().warn(
+            'Unable to call service %s' % service_name)
         return
 
     req = ChangeState.Request()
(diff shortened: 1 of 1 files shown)


clalancette:
Will test it out in a little bit.

lifecycle/src/lifecycle_service_client_py.py
```diff
@@ -24,7 +24,12 @@
 def change_state(lifecycle_node, change_state_args=''):
     node = rclpy.create_node('lc_client_py')
 
-    cli = node.create_client(ChangeState, lifecycle_node + '/change_state')
+    service_name = lifecycle_node + '/change_state'
+    cli = node.create_client(ChangeState, service_name)
+    if not cli.wait_for_service(timeout_sec=5.0):
+        print('Unable to call service', service_name)
```
mikaelarguedas: while this work, we usually use either `%s` or `.format()` to format our printed strings.
Also this entire demo now uses ROS logging so I'd suggest to use logging rather than `print`

Karsten1987:
@mikaelarguedas totally agree. Updated the PR to use logging

lifecycle/src/lifecycle_service_client_py.py
```diff
@@ -49,7 +55,13 @@ def change_state(lifecycle_node, change_state_args=''):
 def get_state(lifecycle_node):
     node = rclpy.create_node('lc_client_py')
 
+    service_name = lifecycle_node + '/get_state'
     cli = node.create_client(GetState, lifecycle_node + '/get_state')
```
mikaelarguedas: Nit: reuse `service_name` rather than `lifecycle_node + '/get_state'`

clalancette:
Sorry this took so long to get to; I finally tested it, and it is a lot friendlier now, so it is a big improvement.

Karsten1987:
ok, taking this as an approve. Merging..!

PR #219, opened 2017-12-11
remove workaround as connext declares dep on libdl now

mikaelarguedas:
connects to ros2/rmw_connext#275

### DIFF
--- a/composition/CMakeLists.txt
+++ b/composition/CMakeLists.txt
@@ -99,10 +99,6 @@ set(libs
   server_component
   client_component)
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-  set(libs
-    "-Wl,--no-as-needed"
-    ${libs}
-    "-Wl,--as-needed")
 endif()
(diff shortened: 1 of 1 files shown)


dhood:
I think this was added specifically for the linktime_composition part of the demo (I imagine otherwise the linker would not bother to link against the various components because they are not coded dependencies in `linktime_composition.cpp`). I don't think it's related to the connext workaround (which was probably redundant by the time this demo was created)

mikaelarguedas:
Thanks for the feedback, yeah that was my guess too and wanted to confirm it would fail without it. As usual this is not ready for review until being placed in the review column.

dhood:
Personally I appreciate input from others as soon as possible if it looks like I'm going in the wrong direction with something, so I figured I'd point it out even if the PR is labelled as "in progress", didn't mean to step on your toes :smile:

mikaelarguedas:
Yeah I understood your intent that's why I thanked the feedback. I'll make it clearer in the future when I'm confirming the need for extra logic and not expecting reviews on it :wink: .

PR #220, opened 2018-01-02
Fix pendulum inertia in physic model.

ThoGit:
After going through the real time programming tutorial, I started reading the code, and realized that there might be an issue with the physic model of the pendulum.
When computing the acceleration caused by the motor: `acceleration = torque / inertia`. Assuming a point mass inertia, the inertia should be: `mass * length^2`


wjwwood:
Thanks @ThoGit for the pr!

Thanks @scpeters for the physics consult 😉.

Here's CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3822/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/961/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3909/)

PR #221, opened 2018-01-09
Fix comments for state callbacks

gbiggs:
Fix the leader comments for two of the state callbacks, which used the wrong state names.

### DIFF
--- a/lifecycle/src/lifecycle_talker.cpp
+++ b/lifecycle/src/lifecycle_talker.cpp
@@ -195,7 +195,7 @@ public:
     return lifecycle_msgs::msg::Transition::TRANSITION_CALLBACK_SUCCESS;
   }
 
-  /// Transition callback for state cleanup
+  /// Transition callback for state cleaningup
   /**
    * on_cleanup callback is being called when the lifecycle node
    * enters the "cleaningup" state.
--- a/lifecycle/src/lifecycle_talker.cpp
+++ b/lifecycle/src/lifecycle_talker.cpp
@@ -165,7 +165,7 @@ public:
     return lifecycle_msgs::msg::Transition::TRANSITION_CALLBACK_SUCCESS;
   }
 
-  /// Transition callback for state activating
+  /// Transition callback for state deactivating
   /**
    * on_deactivate callback is being called when the lifecycle node
    * enters the "deactivating" state.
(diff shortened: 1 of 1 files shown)


lifecycle/src/lifecycle_talker.cpp
```diff
@@ -195,7 +195,7 @@ class LifecycleTalker : public rclcpp_lifecycle::LifecycleNode
     return lifecycle_msgs::msg::Transition::TRANSITION_CALLBACK_SUCCESS;
   }
 
-  /// Transition callback for state deactivating
+  /// Transition callback for state cleanup
```
mikaelarguedas: should this be "cleaningup" instead of "cleanup" ?

wjwwood: I think cleanup reads better myself.

mikaelarguedas: I agree with the fact that it reads better. I figured we wanted to match the name of the transition state defined in the [design document](http://design.ros2.org/articles/node_lifecycle.html) and [in the same docblock below](https://github.com/ros2/demos/pull/221/files#diff-a333d83f31412712a00d05e83145ebb0R201) but am fine with both

wjwwood: Ok, yeah for that it does make sense. It is the "cleaning up" state not the "cleanup of state".

gbiggs: So what's the conclusion? The callback name and the state name should probably match, and ideally they would also match the design document.

mikaelarguedas: :+1: yes the conclusion is to use "cleaningup" to match the design doc and the rest of the code

gbiggs: OK, done. :+1: 

I think the callback name in the demo and the API should also be changed to `on_cleaningup`, but that's a separate issue.

mikaelarguedas: >  think the callback name in the demo and the API should also be changed to on_cleaningup, but that's a separate issue.

I'm going to merge this now but feel free to open a PR to address those.

wjwwood: > I think the callback name in the demo and the API should also be changed to on_cleaningup, but that's a separate issue.

Actually based on the article, I'd say the callback name is right, no? It's this:

capture

State is "CleaningUp" but the transition is called "onCleanup".

mikaelarguedas: :+1:  reading the design doc I have the same reading as @wjwwood

gbiggs: I'm happy with that.

PR #223, opened 2018-01-12
split composition test for better visibility

dirk-thomas:
The first commit adds a missing commit from https://github.com/ros2/demos/pull/96#discussion_r160267778.

The second and third commit split the single test into individual test files in order to get specific feedback which test passes / is flaky / fails.

The fourth commit just adds a repeated message while waiting for the response (to indicate progress).

The builds won't be "bettter" since this is basically a refactoring but it will allow to more precisely identify which tests are flaky / failing:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3865/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/989/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3194/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3943/)


dirk-thomas:
Since it would be very beneficial to have more CI builds across all platforms, build types and repeated for debugging this I will merge this patch even though it didn't have time for any review. If there is any feedback after the fact I will address that afterwards.

PR #232, opened 2018-04-03
Pendulum teleop flakiness reduction

dhood:
As mentioned in https://github.com/ros2/build_cop/issues/94, there is still an unresolved/unidentified issue that causes the teleop test to hang, so this won't resolve flakiness completely. I reason that these changes are still worth merging so that they can be ruled out as the cause for future investigations.

CI linux with connext (only combo tests run for): [Build Status](https://ci.ros2.org/job/ci_linux/4169/)

Repeating 100 times: [Build Status](https://ci.ros2.org/job/ci_linux/4170/) (I have seen on other builds that it's still prone to failure if you keep repeating the tests, but it's better)

### DIFF
--- a/pendulum_control/src/pendulum_demo.cpp
+++ b/pendulum_control/src/pendulum_demo.cpp
@@ -197,8 +197,12 @@ int main(int argc, char * argv[])
       pendulum_controller->on_pendulum_setpoint(msg);
     };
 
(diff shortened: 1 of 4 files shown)


dhood:
Your questions are valid, I had been working under the assumption that it made sense to be that way for the actual pendulum demo (just publishing once then exiting), not sure if that's reasonable or not.

PR #238, opened 2018-06-01
Only keep the last sample in the image tools by default.

clalancette:
Particularly for cam2image, running it without a subscriber
and KEEP_ALL will eventually cause it to exhaust virtual
memory space.  Running it with KEEP_LAST lets it run indefinitely
without growing memory usage.

We saw this problem while just running cam2image for this tutorial https://github.com/ros2/ros2/wiki/Rosbag-with-ROS1-Bridge .  We had left it running for a little while before running the bridge/rosbag, and it would crash after a few minutes.  With this change in place, we can leave it running for as long as we like.

### DIFF
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -77,10 +77,10 @@ int main(int argc, char * argv[])
 
   // Initialize default demo parameters
   bool show_camera = false;
-  size_t depth = 10;
+  size_t depth = rmw_qos_profile_default.depth;
   double freq = 30.0;
-  rmw_qos_reliability_policy_t reliability_policy = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
-  rmw_qos_history_policy_t history_policy = RMW_QOS_POLICY_HISTORY_KEEP_ALL;
+  rmw_qos_reliability_policy_t reliability_policy = rmw_qos_profile_default.reliability;
+  rmw_qos_history_policy_t history_policy = rmw_qos_profile_default.history;
   size_t width = 320;
   size_t height = 240;
   bool burger_mode = false;
@@ -92,10 +92,10 @@ int main(int argc, char * argv[])
   setvbuf(stdout, NULL, _IONBF, BUFSIZ);
 
   // Configure demo parameters with command line options.
-  bool success = parse_command_options(
-    argc, argv, &depth, &reliability_policy, &history_policy, &show_camera, &freq, &width, &height,
-    &burger_mode, &topic);
-  if (!success) {
+  if (!parse_command_options(
+      argc, argv, &depth, &reliability_policy, &history_policy, &show_camera, &freq, &width,
+      &height, &burger_mode, &topic))
+  {
     return 0;
   }
 
--- a/image_tools/src/options.cpp
+++ b/image_tools/src/options.cpp
@@ -59,11 +59,13 @@ bool parse_command_options(
     ss << " -r: Reliability QoS setting:" << std::endl;
     ss << "    0 - best effort" << std::endl;
     ss << "    1 - reliable (default)" << std::endl;
-    ss << " -d: Queue depth. 10 (default)" << std::endl;
+    ss << " -d: Depth of the queue: only honored if used together with “keep last”. " <<
+      "10 (default)" << std::endl;
     ss << " -f: Publish frequency in Hz. 30 (default)" << std::endl;
     ss << " -k: History QoS setting:" << std::endl;
-    ss << "    0 - only keep last sample" << std::endl;
-    ss << "    1 - keep all the samples (default)" << std::endl;
+    ss << "    0 - only store up to N samples, configurable via the queue depth (default)" <<
+      std::endl;
+    ss << "    1 - keep all the samples" << std::endl;
     if (show_camera != nullptr) {
       ss << " -s: Camera stream:" << std::endl;
       ss << "    0 - Do not show the camera stream" << std::endl;
--- a/image_tools/src/options.cpp
+++ b/image_tools/src/options.cpp
@@ -59,7 +59,7 @@ bool parse_command_options(
     ss << " -r: Reliability QoS setting:" << std::endl;
     ss << "    0 - best effort" << std::endl;
     ss << "    1 - reliable (default)" << std::endl;
-    ss << " -d: Depth of the queue: only honored if used together with “keep last”. " <<
+    ss << " -d: Depth of the queue: only honored if used together with 'keep last'. " <<
       "10 (default)" << std::endl;
     ss << " -f: Publish frequency in Hz. 30 (default)" << std::endl;
     ss << " -k: History QoS setting:" << std::endl;
--- a/image_tools/src/showimage.cpp
+++ b/image_tools/src/showimage.cpp
@@ -90,9 +90,9 @@ int main(int argc, char * argv[])
   rclcpp::init(argc, argv);
 
   // Initialize default demo parameters
-  size_t depth = 10;
-  rmw_qos_reliability_policy_t reliability_policy = RMW_QOS_POLICY_RELIABILITY_RELIABLE;
-  rmw_qos_history_policy_t history_policy = RMW_QOS_POLICY_HISTORY_KEEP_ALL;
+  size_t depth = rmw_qos_profile_default.depth;
+  rmw_qos_reliability_policy_t reliability_policy = rmw_qos_profile_default.reliability;
+  rmw_qos_history_policy_t history_policy = rmw_qos_profile_default.history;
   bool show_camera = true;
   std::string topic("image");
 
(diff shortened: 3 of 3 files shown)


clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4563/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1473/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3750/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4647/)

clalancette:
> You're using fastrtps right? I wouldn't have expected the publisher to keep the samples if there are no subscribers (unless we use transient local durability, which we aren't). But if this is the behaviour that we observe, then I agree it makes sense to change the default of the demo.

Fast-RTPS, correct.  Do you think the publisher keeping the samples even if there are no subscribers is a bug we should report to upstream Fast-RTPS?

> The help text of the demo/title of the PR are misleading because keep last will keep up to the queue size, not just 1 sample. I added e8e401a to improve the docs to more closely match https://github.com/ros2/ros2/wiki/About-Quality-of-Service-Settings

:+1: 

> We should delay/keep this change in mind wrt ros2/rmw_fastrtps#202 which is referencing the existing defaults

Good point.  I'll keep this in progress until that is merged, then I'll retest.

dhood:
> Do you think the publisher keeping the samples even if there are no subscribers is a bug we should report to upstream Fast-RTPS?

I'm not sure if the behaviour is within the spec or not, so don't know that it's a bug. We could ask eProsima.

dhood:
> Do you think the publisher keeping the samples even if there are no subscribers is a bug we should report to upstream Fast-RTPS?

@richiprosima did you have any input on if this is the expected behaviour?

dirk-thomas:
> Do you think the publisher keeping the samples even if there are no subscribers is a bug

This is not a bug but expected behavior in DDS. The purpose is to provide all "historical" data to late joining subscribers.

dhood:
even if using durability of volatile as mentioned?

dirk-thomas:
I wouldn't expect it to happen with volatile. Seems like the same thing we discussed earlier: https://github.com/ros2/rmw_fastrtps/issues/68#issuecomment-264559310

richiware:
While publisher was running, was a subscriber running too? In this case, was the subscriber's application killed or stopped neatly?

My thought is that publisher doesn't know the subscriber is not anymore there and is storing the samples until the subscriber acknowledge them.

mikaelarguedas:
I can confirm that this is reproducible without any subscriber running.

Using master, without any other node running, I ran:
`time ros2 run image_tools cam2image -b -x 640 -y 480 -f 400`

After 14 sec, the process is using 4.5GB of RAM and crashes when trying to publish the 5001st sample
```
[INFO] [cam2image]: Publishing image #5000
[INFO] [cam2image]: Publishing image #5001
  what():  failed to publish message: cannot publish data, at /home/mikael/work/ros2/bouncy_ws/src/ros2/rmw_fastrtps/rmw_fastrtps_cpp/src/rmw_publish.cpp:53, at /home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:241

real	0m14.235s
user	0m4.079s
sys	0m2.804s
```

That is expected when we use KEEP_ALL from what I understand ? https://github.com/ros2/rmw_fastrtps/issues/68
Or does @richiware mean that we should not face the 5000 message issue if we use volatile no subscriber was ever matched on that topic?

---

@clalancette The patch should be modified to use the defaults profile from rmw. This problem was faced and worked around in https://github.com/ros2/rmw/pull/82. So I think that this demo should stop hard-coding the history and reliability policy (but keep defining the queue size) and use the default profiles instead

richiware:
Although you use KEEP_ALL, in VOLATILE samples are stored until subscribers acknowledge them. As I see this mechanism is implemented when an ACKNACK message is received.

In your scenario where there aren't subscribers this mechanism is not called and the samples are not removed from the history. This is a bug. I will work on it. Thanks for the reporting.

clalancette:
> Using master, without any other node running, I ran:
> time ros2 run image_tools cam2image -b -x 640 -y 480 -f 400
> 
> After 14 sec, the process is using 4.5GB of RAM and crashes when trying to publish the 5001st sample

@mikaelarguedas Thanks for the additional testing; that is exactly what we saw in testing.

> @clalancette The patch should be modified to use the defaults profile from rmw. This problem was faced and worked around in ros2/rmw#82. So I think that this demo should stop hard-coding the history and reliability policy (but keep defining the queue size) and use the default profiles instead

I think there is some value in having a demo where users can easily play with the "low-level" settings of history and reliability to see what the effects on the system are.  That being said, I think that we should probably use the default profiles as a default, and only modify it if the user passes a flag.  What do you think about that?

> In your scenario where there aren't subscribers this mechanism is not called and the samples are not removed from the history. This is a bug. I will work on it. Thanks for the reporting.

@richiware Thanks for the confirmation.

mikaelarguedas:
>  I think that we should probably use the default profiles as a default, and only modify it if the user passes a flag. What do you think about that?

Yes, that is what I'm suggesting

clalancette:
I re-tested this with the latest Fast-RTPS changes, and this still "leaks" memory if there are no subscribers.  Therefore, I've gone ahead nad rebased this, and changed it to stick with `rmw_qos_profile_default` unless the user explicitly requests something else via a command-line option.

I tested by first running `ros2 run image_tools cam2image -- -b` with no subscribers, and saw that it did not start using more and more memory (because of the KEEP_LAST default).  Then I ran `ros2 run image_tools cam2image -- -b -k 1`, and saw that it used more and more memory (because of the KEEP_ALL requested plus the Fast-RTPS bug).  I also ran it with `ros2 run image_tools cam2image -- -b -d 400`, and saw that it used up memory until it did the 400th publish, at which point it stop using more memory.  So at this point, I think this is doing what we want (and working around the Fast-RTPS bug for now).  I'm going to place it back in review, and new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4658/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1549/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3852/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4733/)

image_tools/src/cam2image.cpp
```diff
@@ -91,36 +88,23 @@ int main(int argc, char * argv[])
...
+  {
     return 0;
   }
 
   // Initialize a ROS 2 node to publish images read from the OpenCV interface to the camera.
   auto node = rclcpp::Node::make_shared("cam2image");
   rclcpp::Logger node_logger = node->get_logger();
 
-  // Set the parameters of the quality of service profile. Initialize as the default profile
-  // and set the QoS parameters specified on the command line.
```
mikaelarguedas: I found these comments valuable even if they come at the cost of a few local variables.
Do you mind keeping `reliability_policy` and friends variables but initialize them with the default profile values instead of the previously hard-coded ones?

clalancette: I did basically this, but just re-used the fields of the structure.  The code looks slightly weird this way, but at least has the documentation that you liked.  I could re-introduce the local variables, but it seems like more of a way to make a mistake than anything else.  Let me know what you think.

mikaelarguedas:
It's not a very big deal but while this allows to keep the comments, all the statements have not effect so I don't know if it's clearer for users.

I'd prefer keeping the temporary variables (initialized with the default values of the qos profile), modify them based on user provided input, and then assign them to the custom qos profile with the appropriate comment.

Then I don't see it as a blocker for getting this merged.

clalancette:
OK, I've rebased now and switched to only using the default arguments for depth, reliability, and history.  I think this should be ready for another review; one more CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4929/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1752/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4091/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4947/)

clalancette:
> Can you look into the failures on windows ? it looks to be due to this change as it doesnt appear on master

Yeah, just looking into it now.  Odd failure, I'm not sure why it is happening.

clalancette:
Got it, Windows didn't like the Unicode quotes.  One more CI on Windows just to confirm that it is fixed:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4951/)

mikaelarguedas:
So this was machine independant after all ? or is there something different on portable than the other machines?

clalancette:
> So this was machine independant after all ? or is there something different on portable than the other machines?

It was not machine dependent.  I made a mistake while trying to reproduce it, and finally found the real problem (the quotes).

mikaelarguedas:
:+1:

clalancette:
CI is passing, and I have two approvals, so I'm going to merge this in.  Thanks!

PR #241, opened 2018-06-08
Declare parameters

sloretz:
This group of changes makes the default behavior to disallow parameters unless they are declared. This PR declares them. Alternatively a boolean could be passed to the node's constructor to allow undeclared parameters.

connects to ros2/rclcpp#495


demo_nodes_cpp/src/parameters/even_parameters_node.cpp
```diff
@@ -26,10 +26,19 @@ class EvenParameterNode : public rclcpp::Node
     // Declare a parameter change request callback
     // This function will enforce that only setting even integer parameters is allowed
     // any other change will be discarded
+    auto existing_callback = this->set_on_parameters_set_callback(nullptr);
```
jacobperron: Is this necessary?
I understand that it's nice to demo how users can handle multiple callbacks, but in this example it could give the wrong impression that a vanilla `Node` has a callback by default (at least I don't think it does). I think it would be better to have a separate demo where multiple callbacks are used, and keep this one succinct.

wjwwood: It does not right now, but if we add one for `use_sim_time`, any code that did not do something like this would subtly break things.

This was an, opinionated, attempt to bring attention to this issue for the users, but it would be simpler if they didn't need to consider it.

Can others comment on what they think?

jacobperron: IMO, it shouldn't be possible for a user to implicitly remove a callback internal to a node (e.g. for `use_sim_time`). If we add a callback for `use_sim_time`, then I think it should be implemented such that users can't accidentally break things by registering their own callback (unless it is explicitly for `use_sim_time`).

dirk-thomas: From an API point of view it would be nicer / safer if the caller user land code wouldn't be responsible for the chaining. That would require though that a list of callback needs to be stored and the semantic for checking them be defined (e.g. each callback would need to return a tri-state ["accept", "reject", "unknown"] and only in the last case the next callback in the chain would be checked and so on).

wjwwood: @jacobperron I don't agree completely. I think that having a special side channel for our features is a bad approach, instead we should use the same interfaces available to the user in order to accomplish what we want (principle of dog-fooding).

I do agree however, that putting the responsibility on the user for chaining is not desirable, but I brought that up for discussion here:

https://github.com/ros2/rclcpp/pull/495#discussion_r275565749

One solution is to have a chain of callbacks, but I didn't want to change that in this pr set because it's already huge and I was trying to avoid (as much as possible) changing existing behavior. The only reason this function is new and the old is deprecated is that the old function doesn't allow you to even get the callback you're overwriting.

Even if we decided to change the way this function works for this pull request (I don't think it's a good idea because of the timelines involved) we'd have to figure out a lot of details and document/test them, e.g.:

- what order a the callbacks called in?
- what happens when a callback returns a result that is unsuccessful, do you continue and && the results or stop on the first rejection, and if the former how do you combine the error messages? 
- how can you remove a callback once added?

And some others that I have in my head after thinking about taking this on in this pull request a few weeks ago.

dirk-thomas: Maybe we get this in an create a ticket for improving the API in the future (likely post Dashing) with the points brought up in the discussion?

jacobperron: I think a chain of callbacks solution makes sense (and in a later PR). Going back to my original question, I was wondering if we should chain the existing callback in this demo (which is always `nullptr`). If we add a callback for `use_sim_time` later then that seems like a good time to change the behavior, otherwise we risk subtly breaking things for users who are not adhering to this chaining pattern. At the very least, it's probably worth putting an additional comment explaining why it's important to do the chaining here.

wjwwood: What should I do here? I'm happy to either remove this, with or without changing the implementation to allow for more than one vector at a time, or leave it.

I still think exposing this inconvenient reality to the user is a good idea, but honestly I don't think it's that important one way or the other, especially since I'd like to see the function change yet again in the future to support multiple callbacks simultaneously.

jacobperron: Since it will change in the future, I'm okay with leaving it as-is.

tfoote: +1 for leaving it as is with an enhancement ticket to support adding callbacks that then will automatically be sequenced.

PR #242, opened 2018-06-11
 Fix potiential nullptr dereferenced issue in demo_nodes_cpp

testkit:
Need check if it's nullptr before do std::string() convert.

### DIFF
--- a/demo_nodes_cpp/src/services/add_two_ints_client.cpp
+++ b/demo_nodes_cpp/src/services/add_two_ints_client.cpp
@@ -65,9 +65,11 @@ int main(int argc, char ** argv)
   }
 
   auto topic = std::string("add_two_ints");
-  if (rcutils_cli_option_exist(argv, argv + argc, "-s")) {
-    topic = std::string(rcutils_cli_get_option(argv, argv + argc, "-s"));
+  char * cli_option = rcutils_cli_get_option(argv, argv + argc, "-s");
+  if (nullptr != cli_option) {
+      topic = std::string(cli_option);
   }
+
   auto client = node->create_client<example_interfaces::srv::AddTwoInts>(topic);
 
   auto request = std::make_shared<example_interfaces::srv::AddTwoInts::Request>();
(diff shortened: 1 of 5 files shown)


demo_nodes_cpp/src/services/add_two_ints_client.cpp
```diff
@@ -65,9 +65,15 @@ int main(int argc, char ** argv)
   }
 
   auto topic = std::string("add_two_ints");
+
   if (rcutils_cli_option_exist(argv, argv + argc, "-s")) {
```
dhood: we might as well remove this check now that we do not assume that `rcutils_cli_get_option` will return a valid string.

demo_nodes_cpp/src/services/add_two_ints_client.cpp
```diff
@@ -65,9 +65,15 @@ int main(int argc, char ** argv)
...
-    topic = std::string(rcutils_cli_get_option(argv, argv + argc, "-s"));
+    char * cli_option = rcutils_cli_get_option(argv, argv + argc, "-s");
+
+    if (NULL != cli_option) {
```
dhood: please use `nullptr` instead of `NULL` in C++

demo_nodes_cpp/src/services/add_two_ints_client.cpp
```diff
@@ -65,9 +65,15 @@ int main(int argc, char ** argv)
   }
 
   auto topic = std::string("add_two_ints");
+
   if (rcutils_cli_option_exist(argv, argv + argc, "-s")) {
-    topic = std::string(rcutils_cli_get_option(argv, argv + argc, "-s"));
+    char * cli_option = rcutils_cli_get_option(argv, argv + argc, "-s");
+
```
dhood: please remove the blank space between these lines as they are directly related

testkit:
@dhood,  thanks for your review. PR updated as below:
  Edit based on comments about nullptr, blank space, redundent rcutils_cli_option_exist() calling.
  Fix the similar bug in topics/listener.cpp, topics/talker.cpp

dhood:
I fixed the style errors. @testkit in the future please be sure to run tests for your code changes because we also have tests that check that the code complies with our style guidelines.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4739/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1607/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3927/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4799/)

testkit:
@dhood , noted and thanks!

PR #244, opened 2018-06-14
add talker_listener.launch.py new-style launch file

wjwwood:
Connects to ros2/launch#76

### DIFF
--- a/demo_nodes_cpp/CMakeLists.txt
+++ b/demo_nodes_cpp/CMakeLists.txt
@@ -124,4 +124,10 @@ if(BUILD_TESTING)
   call_for_each_rmw_implementation(tests)
 endif()
 
+# Install launch files.
+install(DIRECTORY
+  launch
+  DESTINATION share/${PROJECT_NAME}/
+)
+
 ament_package()
--- a/demo_nodes_cpp/launch/services/add_two_ints.launch.py
+++ b/demo_nodes_cpp/launch/services/add_two_ints.launch.py
@@ -1,0 +1,36 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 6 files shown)


wjwwood:
I'd like to expand this to include launch files for everything, but after we get to review.

wjwwood:
I expanded the launch files to include all the topics and service demos in `demo_nodes_cpp`. If anyone has time to review these quickly, I'd appreciate it!

dirk-thomas:
Should the launch files have the executable bit set?

dirk-thomas:
Nvm - the can't be invoked in their own... Ignore the noise.

wjwwood:
I'll look into the hang.

dhood:
think this should add a dependency on `launch_ros` in the package.xml yeah?

wjwwood:
> think this should add a dependency on launch_ros in the package.xml yeah?

Not sure, I guess you could, but a similar thing comes up where you wonder if every package in ROS 1 that has a launch file should depend on `roslaunch`.

If we decide to add a dependency, then we might as well add `ros2launch` in addition to (or instead of) `launch_ros`. Because those launch files cannot be run on their own.

dhood:
I think adding a `ros2launch` dependency is a step further than necessary, because even though they're not usable on their own, it doesn't have to be `ros2launch` that uses them (could theoretically be anyone).

But I take your point about having all ROS 1 packages depend on roslaunch. In that sense I think it's reasonable that only packages that are actually launching launch files (e.g. in tests) depend on `launch_ros` themselves.

dirk-thomas:
Since the launch files do import `launch_ros` I think the packages should have an `exec_depend` on `launch_ros`.

Regarding the dependency on `ros2launch`: while I can see the convenience if someone would only install that package how it would help I wouldn't add it since `ros2launch` is just one way these launch files can be used. As an analogy: if a package provides a component (a node compiled into a shared library) I wouldn't expect a dependency on something like `ros2run-component` (not that we have that - just as an example).

wjwwood:
https://github.com/ros2/launch/issues/80

PR #246, opened 2018-06-19
Topic monitor new launch

dhood:
I noticed that scripts being run from within launch were receiving two SIGINTs (once from ctrl-c, and once from (legacy) launch [sending one](https://github.com/ros2/launch/blob/fb6cbe4bd79b494f2ee85808f6a580a1a4b45111/launch/launch/legacy/launcher.py#L238)). I was going to ignore signals [after the first KeyboardInterrupt](https://github.com/ros2/demos/blob/ef033c9b75d34b9745ca82908f4aaea1bd0d568f/topic_monitor/topic_monitor/scripts/data_publisher.py#L118) in the script that gets run by launch, but figured instead I'd swap to new-style launch files.

Note: I'm using `ExecutableProcess` instead of `Node` because I am launching another python script, not a node (i.e. it wouldn't accept __ns arguments).


topic_monitor/topic_monitor/launch/reliability_demo.launch.py
```diff
@@ -0,0 +1,30 @@
...
+"""Launch data publishers with different reliability configruations."""
+
+from launch import LaunchDescription
+import launch.actions
+from ros2run.api import get_executable_path
+
+
+def generate_launch_description():
+    executable = get_executable_path(
+        package_name='topic_monitor', executable_name='data_publisher')
```
wjwwood: I recommend using the `launch_ros.substitutions.ExecutableInPackage` here.

dhood:
Symlink installs were having issues with `data_files` referencing launch files inside the source directory of the python package (the launch files would end up symlinked to themselves and not be installable; I originally misdiagnosed it as https://github.com/ros2/launch/issues/83).

Now I have moved the launch files outside the `topic_monitor` subdirectory (as with the other resource files), but it means that they don't get symlinked. If people have suggestions on the right way to address this we can fix it in a followup PR.

dhood:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4765/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1630/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4822/)

PR #247, opened 2018-06-20
Switch order of arguments back to roscpp-like style

esteve:
Connects to ros2/rclcpp#503


nuclearsandwich:
Since the connecting PR is closed I am going to close this one as well.

PR #258, opened 2018-06-22
check that resolution is at least as big as burger

### LINKED ISSUE #254: cam2image crashes when specifying -x -y and -b
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - binaries
- Version or commit hash:
  - https://ci.ros2.org/view/packaging/job/packaging_windows/1042/
- DDS implementation:
  - Fast-RTPS
  - N/A

#### Steps to reproduce issue

```
ros2 run image_tools cam2image -b -x 16 -y 16
```

#### Expected behavior
Either a smaller image would be published or a message would be printed saying resizing the generated image is not supported.

#### Actual behavior
```
>ros2 run image_tools cam2image -b -x 16 -y 16
[INFO] [cam2image]: Publishing data on topic 'image'
OpenCV(3.4.1) Error: Assertion failed (0 <= roi.x && 0 <= roi.width && roi.x + roi.width <= m.cols && 0 <= roi.y && 0 <= roi.height && roi.y + roi.height <= m.rows) in cv::Mat::Mat, file C:\Users\osrf\opencv3-build\opencv\modules\core\src\matrix.cpp, line 464
```

#### Additional information

dirk-thomas:
Fixes #254.

The first check ensures that the resolution is at least big enough to contain the burger template.

Additionally the initialization avoids a division by zero and the "bouncing" sacrifices "accuracy" over staying in bound in very constraint sizes.

### DIFF
--- a/image_tools/src/burger.cpp
+++ b/image_tools/src/burger.cpp
@@ -95,6 +95,11 @@ cv::Mat & Burger::render_burger(size_t width, size_t height)
 {
   int width_i = static_cast<int>(width);
   int height_i = static_cast<int>(height);
+  if (width_i < burger_template.size().width || height_i < burger_template.size().height) {
+    std::string msg = "Target resolution must be at least the burger size (" +
+      std::to_string(burger_template.size().width) + " x " + std::to_string(burger_template.size().height) + ")";
+    throw std::runtime_error(msg.c_str());
+  }
   if (burger_buf.size().width != width_i || burger_buf.size().height != height_i) {
     int num_burgers = rand() % 10 + 2;  // NOLINT
     x.resize(num_burgers);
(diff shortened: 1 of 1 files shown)


image_tools/src/burger.cpp
```diff
@@ -122,11 +135,19 @@ cv::Mat & Burger::render_burger(size_t width, size_t height)
     // bounce as needed
     if (x[b] < 0 || x[b] > width_i - burger_template.size().width - 1) {
       x_inc[b] *= -1;
-      x[b] += 2 * x_inc[b];
+      if (x[b] < 0) {
+        x[b] = 0;
+      } else {
+        x[b] = width_i - burger_template.size().width;
```
sloretz: Looks like `x_inc[b]` is unused?

dirk-thomas: Correct it just truncates the movement at the border. It had already been moved in line 133.

PR #266, opened 2018-07-05
Fix no return code for main() in several files

testkit:
return code is missing in main(), add "return 0;"


dirk-thomas:
Thanks for adding the explicit return value.

PR #273, opened 2018-08-02
add example of registering custom parameter validation callbacks

mikaelarguedas:
fixes https://github.com/ros2/examples/issues/164

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4994/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1806/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4150/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5007/)

### DIFF
--- a/demo_nodes_cpp/src/parameters/even_parameters_node.cpp
+++ b/demo_nodes_cpp/src/parameters/even_parameters_node.cpp
@@ -1,0 +1,86 @@
+// Copyright 2018 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <memory>
+#include <vector>
+
+#include "rclcpp/rclcpp.hpp"
+
+class EvenParameterNode : public rclcpp::Node
+{
+public:
+  EvenParameterNode()
+  : Node("even_parameters_node")
+  {
+    // Declare a parameter change request callback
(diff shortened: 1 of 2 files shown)


demo_nodes_cpp/src/parameters/even_parameters_node.cpp
```diff
@@ -0,0 +1,86 @@
...
+              RCLCPP_INFO(this->get_logger(),
+                "Requested value '%d' for parameter '%s' is not an even number:"
+                " rejecting change...",
+                parameter.as_int(),
+                parameter.get_name().c_str()
+              );
+              result.successful = false;
+            } else {
+              RCLCPP_INFO(this->get_logger(),
+                "parameter '%s' has changed and is now %s",
```
wjwwood: Might want to quote the second `%s`.

wjwwood: Or use a `: %s`

mikaelarguedas:
> The only other thing I might look for is an example that uses std::bind to pass an arbitrary argument into the param_change_callback along with the parameters. 

I think we would need another story to justify passing extra arguments to this callback as currently its only purpose is to check for parity and I have a hard time finding a variable to pass in that would make sense.

demo_nodes_cpp/src/parameters/even_parameters_node.cpp
mikaelarguedas: Good point, I purposefully avoided the quoting as I was afraid it could confuse users into thinking that a string value was set and not an integer.
I went for the 2nd approach in https://github.com/ros2/demos/pull/273/commits/61098803d7856a20f48032da4cd1bbac27a81466

mikaelarguedas:
@clalancette Posting new comment instead of editing to sent notification:

> I think we would need another story to justify passing extra arguments to this callback

I'm open to suggestions for this.
If we don't find anything compelling by tomorrow I'm planning on merging this as is.

clalancette:
> If we don't find anything compelling by tomorrow I'm planning on merging this as is.
> Merge state

Nothing really compelling comes to mind, so I'd say go ahead and merge.  We can always revisit later.

mikaelarguedas:
Sounds good, thanks for the review!

PR #275, opened 2018-09-06
change timer with period of 0s to 1ms

dirk-thomas:
Related to ros2/rcl#291.

A timer with a period of 0s is kind of weird. It implies an infinite frequency.


wjwwood:
I don't think this is the right way to resolve this. There's already a TODO to have a "call soon" like function on the node, at which point we could make `0s` an invalid value for timers, until then `1ms`, while similar, is not the same thing.

If you truly believe a period of zero makes no sense, then it should not be allowed as an argument to the create timer function.

PR #287, opened 2018-11-25
Added serialized listener demo for python

josephduchesne:
Connects to ros2/rclpy#242

This was added at the suggestion of @Karsten1987 for ros2/rclpy/pull/242

This is a python analog of listener_serialized_message.cpp in demo_nodes_cpp.

For it to work, you need to be on the rclpy branch in my pull request, or that PR has to be merged in.

As a result, this should not be merged until ros2/rclpy/pull/242 is.


demo_nodes_py/demo_nodes_py/topics/listener_serialized.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
```
Karsten1987: nitpick: year

demo_nodes_py/demo_nodes_py/topics/listener_serialized.py
```diff
@@ -0,0 +1,51 @@
...
+class SerializedSubscriber(Node):
+
+    def __init__(self):
+        super().__init__('serialized_subscriber')
+        self.subscription = self.create_subscription(
+            String,
+            'chatter',
+            self.listener_callback,
+            raw=True)  # We're subscribing to the serialized bytes, not std_msgs.msg.String
+        self.subscription  # prevent unused variable warning
```
Karsten1987: it looks like we don't need that in our other python demos. 
https://github.com/ros2/demos/blob/master/demo_nodes_py/demo_nodes_py/topics/listener.py#L25
Where did you encounter this warning?

Karsten1987:
@josephduchesne I hope you don't mind. I edited your comment to connect this PR to your `rclpy` one in a way that the waffle board picks it up.

demo_nodes_py/demo_nodes_py/topics/listener_serialized.py
josephduchesne: Ah I actually copied this from the rclpy examples. It can probably be removed.

PR #288, opened 2018-12-01
Prevent frame going out of scope when converting RGB -> BGR

sgvandijk:
In `showimage`, when an image is encoded in RGB, it is converted into a BGR image. However, the converted frame is deallocated when the `cv:Mat` referencing it, `frame2`, goes out of scope, resulting in a segmentation fault when trying to show it.

This PR fixes this by reusing the original `cv::Mat` in `cv::cvtColor`, which is safe, and stays in scope until the end of the function.

### DIFF
--- a/image_tools/src/showimage.cpp
+++ b/image_tools/src/showimage.cpp
@@ -67,14 +67,11 @@ void show_image(
       msg->height, msg->width, encoding2mat_type(msg->encoding),
       const_cast<unsigned char *>(msg->data.data()), msg->step);
 
-    CvMat cvframe;
     if (msg->encoding == "rgb8") {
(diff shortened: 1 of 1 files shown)


wjwwood:
Thanks @sgvandijk! Here's CI and I'll merge if those come back greed:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5824/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2389/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4815/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5719/)

PR #296, opened 2018-12-12
Do not show freq parameter on help when not necessary

gonzodepedro:
Do not show frequency parameter on image_tools when not necessary. Relate ros2/ros2_documentation#41

### DIFF
--- a/image_tools/src/options.cpp
+++ b/image_tools/src/options.cpp
@@ -61,7 +61,9 @@ bool parse_command_options(
     ss << "    1 - reliable (default)" << std::endl;
     ss << " -d: Depth of the queue: only honored if used together with 'keep last'. " <<
       "10 (default)" << std::endl;
-    ss << " -f: Publish frequency in Hz. 30 (default)" << std::endl;
+    if (freq != nullptr) {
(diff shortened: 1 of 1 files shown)


image_tools/src/options.cpp
```diff
@@ -61,7 +61,9 @@ bool parse_command_options(
     ss << "    1 - reliable (default)" << std::endl;
     ss << " -d: Depth of the queue: only honored if used together with 'keep last'. " <<
       "10 (default)" << std::endl;
-    ss << " -f: Publish frequency in Hz. 30 (default)" << std::endl;
+    if( freq != nullptr ){
```
dirk-thomas: I would expect this line to fail the linters. Please run the unit tests of the package locally to check.

gonzodepedro: Got it. Fixed now

dirk-thomas: I don't see any change.

gonzodepedro: I'm very sorry. Now it should be fixed.

dirk-thomas:
Thank you for the improvement.

PR #300, opened 2018-12-13
add readme.rst

Karsten1987:
This is referring to https://github.com/ros2/ros2_documentation/pull/47#issuecomment-446795736

The documentation of the lifecycle (in fact every demo) should be contained within the package and not duplicated in the `ros2_documentation` package.

### DIFF
--- a/lifecycle/README.rst
+++ b/lifecycle/README.rst
@@ -198,17 +198,22 @@ Every child of LifecycleNodes have a set of callbacks provided.
 These callbacks go along with the applied state machine attached to it.
 These callbacks are:
 
+.. code-block:: c
 
-* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-    on_configure(const rclcpp_lifecycle::State & previous_state)``
-* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-    on_activate(const rclcpp_lifecycle::State & previous_state)``
-* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-    on_deactivate(const rclcpp_lifecycle::State & previous_state)``
-* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-    on_cleanup(const rclcpp_lifecycle::State & previous_state)``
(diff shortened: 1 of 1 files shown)


lifecycle/README.rst
```diff
@@ -0,0 +1,343 @@
...
+* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+    on_configure(const rclcpp_lifecycle::State & previous_state)``
+* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+    on_activate(const rclcpp_lifecycle::State & previous_state)``
+* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+    on_deactivate(const rclcpp_lifecycle::State & previous_state)``
+* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+    on_cleanup(const rclcpp_lifecycle::State & previous_state)``
+* ``rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+    on_shutdown(const rclcpp_lifecycle::State & previous_state)``
```
sloretz: The functions look a bit weird on github's preview, but the bouncy version looks fine https://github.com/ros2/demos/blob/bouncy/lifecycle/README.rst

Karsten1987: should be addressed now

lifecycle/README.rst
```diff
@@ -0,0 +1,353 @@
...
+   [lc_talker] Lifecycle publisher is active. Publishing: [Lifecycle HelloWorld #12]
+   ...
+
+The lifecycle listener receives the same set of notifications as before.
+Lifecycle talker changed its state from inactive to active.
+
+.. code-block:: bash
+
+   [lc_listener] notify callback: Transition from state unconfigured to configuring
+   [lc_listener] notify callback: Transition from state configuring to inactive
```
sloretz: I see two more transitions after these before the listener receives data

```
[INFO] [lc_listener]: notify callback: Transition from state inactive to activating
[INFO] [lc_listener]: notify callback: Transition from state activating to active
```

Karsten1987: good catch. That was a copy paste error.

PR #306, opened 2019-01-10
Update for OpenCV v4.0 compatibility

jacobperron:


image_tools/src/burger.cpp
```diff
@@ -124,7 +124,7 @@ cv::Mat & Burger::render_burger(size_t width, size_t height)
     }
     burger_buf = cv::Mat(height_i, width_i, CV_8UC3);
   }
-  burger_buf = cvScalar(0, 0, 0);
+  burger_buf = cv::Scalar_<int>(0, 0, 0);
```
wjwwood: Is there any reason you're using `int` specifically rather than `double`, i.e. `cv::Scalar`? I assume there is, but I'm just curious.

jacobperron: Oops, no. I think this should just be `cv::Scalar` :grimacing: 

Fixed in 2877adb

jacobperron:
Packages up to image_tools:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6018/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2524/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4968/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5892/)

PR #311, opened 2019-02-11
Drops legacy launch API usage.

hidmic:
Connected to ros2/launch#159.

### DIFF
--- a/composition/package.xml
+++ b/composition/package.xml
@@ -28,7 +28,6 @@
   <test_depend>ament_cmake_pytest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
-  <test_depend>launch</test_depend>
   <test_depend>launch_testing</test_depend>
   <test_depend>rmw_implementation_cmake</test_depend>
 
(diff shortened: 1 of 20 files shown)


composition/package.xml
```diff
@@ -28,7 +28,6 @@
   <test_depend>ament_cmake_pytest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
-  <test_depend>launch</test_depend>
```
dirk-thomas: Why was this dependency removed? Python modules from that package are clearly imported in the code below.

mjcarroll: Yes, it clearly still needs that dep.

mjcarroll: Though launch_testing includes `launch` as an `exec_depend`.  Is that sufficient in this case?

dirk-thomas: You never want to rely on transitive dependencies. If you use something directly you should declare it as a dependency. What if `launch_testing` would be refactored into a standalone package? Then this package would fail.

hidmic: Oops, my bad. I _think_ I removed them during a previous refactor, when `launch_testing` was the only thing needed for testing. I'll get them back ASAP.

PR #314, opened 2019-03-08
Example of using Deadline, Liveliness, and Lifespan QoS in Client

mm318:
As part of implementation for Deadline, Liveliness, and Lifespan QoS. The design is at ros2/design#212

The changes are examples of how a user would create subscribers and publishers in their client-side application code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,3 +11,8 @@ be under the Apache 2 License, as dictated by that
    the terms of any separate license agreement you may have executed
    with Licensor regarding such Contributions.
 ~~~
+
+Contributors must sign-off each commit by adding a `Signed-off-by: ...`
+line to commit messages to certify that they have the right to submit
+the code they are contributing to the project according to the
+[Developer Certificate of Origin (DCO)](https://developercertificate.org/).


demo_nodes_cpp/src/topics/allocator_tutorial.cpp
```diff
@@ -161,6 +161,17 @@ int main(int argc, char ** argv)
     node = rclcpp::Node::make_shared("allocator_tutorial", options);
   }
 
+  auto event_callback = [](rclcpp::ResourceStatusEvent event_type) -> void
+    {
+      if (rclcpp::ResourceStatusEvent::DEADLINE_MISSED == event_type) {
+        printf("deadline event occured");
```
wjwwood: As we spoke about in our meeting, I think it would be best to have separate callbacks for each type of event, so that you could pass additional information about that event clearly as arguments to the callback. For example, here it would be good to know which instance/publisher or instance/subscription had a deadline missed. For the subscription/publisher, you can perhaps automatically associate which one it was b

mm318: Yep, we are currently working on that.

wjwwood: Ok, should I hold off reviewing for now? Or do you want more early feedback?

wjwwood: (I was going to do a full review, and accidentally clicked "add single comment" mid typing 🤦‍♂️

mm318: We would like more early feedback. We believe that rcl (https://github.com/ros2/rcl/pull/396) and rmw (https://github.com/ros2/rmw/pull/167) interfaces are relatively final.

rclcpp (https://github.com/ros2/rclcpp/pull/655) is probably the most in flux, but for now can you take a look at how we're planning to have PublisherBase and SubscriptionBase to implement the Waitable interface and removing a bunch of vectors of rcl handles in AllocatorMemoryStrategy?

dummy_robot/dummy_map_server/src/dummy_map_server.cpp
```diff
@@ -33,7 +33,7 @@ int main(int argc, char * argv[])
   latched_qos.depth = 1;
   latched_qos.durability = RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL;
   auto map_pub = node->create_publisher<nav_msgs::msg::OccupancyGrid>(
-    "map", latched_qos);
+    "map", nullptr, rclcpp::PublisherOptions<>().publisher_qos_profile(latched_qos));
```
mm318: TODO: Improve usability of `PublisherOptions` and `SubscriptionOptions`.

mm318:
Hi @wjwwood, I've updated the sample code to show the use of separate callbacks for each type of event.

image_tools/src/cam2image.cpp
```diff
@@ -105,26 +105,26 @@ int main(int argc, char * argv[])
...
   // KEEP_LAST enforces a limit on the number of messages that are saved, specified by the "depth"
   // parameter.
-  custom_camera_qos_profile.history = history_policy;
+  custom_camera_qos_profile.qos_profile().history = history_policy;
 
   RCLCPP_INFO(node_logger, "Publishing data on topic '%s'", topic.c_str());
   // Create the image publisher with our custom QoS profile.
-  auto pub = node->create_publisher<sensor_msgs::msg::Image>(
-    topic, custom_camera_qos_profile);
+  auto pub = node->create_publisher<sensor_msgs::msg::Image>(topic, nullptr,
```
nburek: If something is getting passed in as a `nullptr` should it also be moved to the Options class?

image_tools/src/cam2image.cpp
```diff
@@ -105,26 +105,26 @@ int main(int argc, char * argv[])
 
   // Set the parameters of the quality of service profile. Initialize as the default profile
   // and set the QoS parameters specified on the command line.
-  rmw_qos_profile_t custom_camera_qos_profile = rmw_qos_profile_default;
+  rclcpp::PublisherOptions<> custom_camera_qos_profile;
```
nburek: Rename variable as it is no longer just qos settings.

image_tools/src/cam2image.cpp
```diff
@@ -139,11 +139,11 @@ int main(int argc, char * argv[])
     };
 
   // Set the QoS profile for the subscription to the flip message.
-  rmw_qos_profile_t custom_flip_qos_profile = rmw_qos_profile_sensor_data;
-  custom_flip_qos_profile.depth = 10;
+  rclcpp::SubscriptionOptions<> custom_flip_qos_profile(rmw_qos_profile_sensor_data);
```
nburek: Update variable name here and all the others below too.

Issue #316, opened 2019-03-20
[rclcpp] Unhandled RCLError exception on keyboard interrupt


Karsten1987:
## Bug report

similar to #305 in python.

**Required Info:**

- Operating System:
  - experienced on OSX, expecting similar behavior on linux
- Installation type:
  - source
- Version or commit hash:
  - master
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue
Start `ros2 run demo_nodes_cpp add_two_ints_client_async`

`crtl-c` it 

#### Expected behavior
Client shuts down correctly.

#### Actual behavior

```
 ➭ ros2 run demo_nodes_cpp add_two_ints_client_async
[INFO] [add_two_ints_client]: service not available, waiting again...
[INFO] [add_two_ints_client]: service not available, waiting again...
[INFO] [add_two_ints_client]: service not available, waiting again...
^C[INFO] [rclcpp]: signal_handler(signal_value=2)
[ERROR] [add_two_ints_client]: Interrupted while waiting for the service. Exiting.

>>> [rcutils|error_handling.c:106] rcutils_set_error_state()
This error state is being overwritten:

  'rcl node's context is invalid, at /Users/karsten/workspace/osrf/ros2_full/src/ros2/rcl/rcl/src/rcl/node.c:471'

with this new error message:

  'the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at /Users/karsten/workspace/osrf/ros2_full/src/ros2/rcl/rcl/src/rcl/guard_condition.c:69'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
libc++abi.dylib: terminating with uncaught exception of type rclcpp::exceptions::RCLError: Failed to create interrupt guard condition in Executor constructor: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at /Users/karsten/workspace/osrf/ros2_full/src/ros2/rcl/rcl/src/rcl/guard_condition.c:69
```

#### Implementation considerations
The future handling is not correct as far as I can tell. I'll work on a PR for it.

clalancette:
Since we don't support macOS anymore, and this seems to work fine in Linux, closing.

PR #317, opened 2019-03-20
spin until future complete for async service client

Karsten1987:
fixes ros2/demos#316 

When hitting `ctrl-c`, the node exits cleanly now:
```
[INFO] [add_two_ints_client]: service not available, waiting again...
[INFO] [add_two_ints_client]: service not available, waiting again...
^C[INFO] [rclcpp]: signal_handler(signal_value=2)
[ERROR] [add_two_ints_client]: Interrupted while waiting for the service. Exiting.
```

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/demo_nodes_cpp/src/services/add_two_ints_client_async.cpp
+++ b/demo_nodes_cpp/src/services/add_two_ints_client_async.cpp
@@ -24,6 +24,8 @@
 #include "example_interfaces/srv/add_two_ints.hpp"
 
 using namespace std::chrono_literals;
+using ServiceResponseFuture =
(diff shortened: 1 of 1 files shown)


jacobperron:
I think it might make for a more compelling demo if the spinning on the future was encapsulated in the class, although it's a bit more involved. For example, in ros2/examples#230 I have class methods "spin_until_goal_complete" (client) and "spin" (server) for encapsulating the logic related to the futures.

nuclearsandwich:
@Karsten1987 with the composition transition underway what is the status of this PR?

jacobperron:
We can try to address it in #325 (or post) to avoid conflicts.

@skucheria FYI

PR #319, opened 2019-03-26
Update for new rclcpp_components

mjcarroll:
Modify the demos to work with the new composition mechanism from rclcpp_components.

Connects to ros2/rclcpp#665


wjwwood:
Does this need to be updated at all? https://index.ros.org/doc/ros2/Tutorials/Composition/

mjcarroll:
@wjwwood Yes docs need to be updated, I'm going to do it in a follow-up PR.

composition/include/composition/client_component.hpp
```diff
@@ -26,7 +26,7 @@ class Client : public rclcpp::Node
 {
 public:
   COMPOSITION_PUBLIC
-  Client();
+  explicit Client(rclcpp::NodeOptions options);
```
dirk-thomas: Should this be a const reference as in the Node API?

dirk-thomas: See #323.

PR #324, opened 2019-04-23
Add basic composition launch demo

mjcarroll:

PR #326, opened 2019-04-25
rename launch file, avoid redundant default actions, and output to screen

wjwwood:
Follow up of https://github.com/ros2/demos/pull/324

### DIFF
--- a/composition/launch/composition_demo.launch.py
+++ b/composition/launch/composition_demo.launch.py
@@ -14,15 +14,13 @@
 
 """Launch a talker and a listener in a component container."""
 
-from launch_ros import get_default_launch_description
+import launch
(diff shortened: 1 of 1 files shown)


composition/launch/composition_demo.launch.py
```diff
@@ -37,8 +35,8 @@ def generate_launch_description():
                     package='composition',
                     node_plugin='composition::Listener',
                     node_name='listener')
-            ]
+            ],
+            output='screen',
```
wjwwood: I assume printing to screen is what you want by default? Maybe not, since you can force it out manually with `ros2 launch -a`?

mjcarroll: Yes, printing as default is what I was going for.

PR #327, opened 2019-04-29
Corrected publish calls with shared_ptr signature

ivanpauno:
Connects to ros2/rclcpp#709

### DIFF
--- a/image_tools/src/cam2image.cpp
+++ b/image_tools/src/cam2image.cpp
@@ -57,17 +57,17 @@ mat_type2encoding(int mat_type)
  * \param[out] Allocated shared pointer for the ROS Image message.
  */
 void convert_frame_to_message(
-  const cv::Mat & frame, size_t frame_id, sensor_msgs::msg::Image::SharedPtr msg)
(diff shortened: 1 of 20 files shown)


dirk-thomas:
In many cases same comment as here: https://github.com/ros2/rclcpp/pull/709#discussion_r279500228

wjwwood:
Actually, should we prefer `unique_ptr`? In the case that intra-process is enabled, it might be better to prefer `unique_ptr`, but if you're only doing inter-process then the stack reference is probably preferable.

Thoughts?

image_tools/src/cam2image.cpp
```diff
@@ -168,14 +169,13 @@ int main(int argc, char * argv[])
...
 
-  // Initialize a shared pointer to an Image message.
-  auto msg = std::make_shared<sensor_msgs::msg::Image>();
-  msg->is_bigendian = false;
-
   size_t i = 1;
 
   // Our main event loop will spin until the user presses CTRL-C to exit.
   while (rclcpp::ok()) {
+    // Initialize a shared pointer to an Image message.
```
sloretz: nit, `shared` -> `unique`

PR #331, opened 2019-05-03
turtlesim for ROS2

routiful:
I have been ported turtlesim node(including tutorials) to ROS2.
It works same as ROS but has some differences in tutorials.

1. Add some information when mimic node is started
(https://github.com/ROBOTIS-Platform/demos/blob/a993136c60d2f9cde7b6f840e2e7bc3c8f735b66/turtlesim/tutorials/mimic.cpp#L15)

1. Add argument to add namespace in mimic node
(https://github.com/ROBOTIS-Platform/demos/blob/a993136c60d2f9cde7b6f840e2e7bc3c8f735b66/turtlesim/tutorials/mimic.cpp#L64)

1. Add stop signal in turtle_teleop node
(https://github.com/ROBOTIS-Platform/demos/blob/a993136c60d2f9cde7b6f840e2e7bc3c8f735b66/turtlesim/tutorials/teleop_turtle_key.cpp#L116)

Please review this and feel free to comment for anything.


dirk-thomas:
Thank you working on this. In general we recommend to contribute the ROS 2 specific change in the upstream ROS 1 repo instead of forking or as in this case even duplicating the sources without any connection.

I have created a [dashing-devel](https://github.com/ros/ros_tutorials/tree/dashing-devel) branch in the `ros_tutorials` repo which you can target for the porting effort. Please try to keep the diff as minimal as possible in order to keep porting patches between the branches easy.

robotpilot:
Thanks, @dirk-thomas ,

But we have to consider one more thing. Other examples of the [ros_tutorials](https://github.com/ros/ros_tutorials/tree/dashing-devel) package (`roscpp_tutorials`, `rospy_tutorials`) are very similar to [ros2/demos](https://github.com/ros2/demos) packages and have the same purpose. I thought we don't need to migrate these packages(`roscpp_tutorials`, `rospy_tutorials`), so we committed to this ros2/demo package. What do you think of this issue?

mlanting:
I've also made a port of turtlesim which I started a few days before I saw this PR come up. I wasn't sure either which repo I was going to try to put it in, but it seems you've beaten me to it anyway. I've been doing my work off of the `ros_tutorials` repository, but my thoughts about where a final version should live were along the same lines as @robotpilot - the rest of this repository already seems to be pretty equivalent to `ros_tutorials` in terms of content and it would be unintuitive for people new to ROS2 to have to go to the ROS 1 tutorials package to find it.

There are already [ros2/tutorials](https://github.com/ros2/tutorials), [ros2/examples](https://github.com/ros2/examples), and [ros2/demos](https://github.com/ros2/demos) which is a bit much to navigate.

dirk-thomas:
> Other examples of the [ros_tutorials](https://github.com/ros/ros_tutorials/tree/dashing-devel) package (`roscpp_tutorials`, `rospy_tutorials`) are very similar to [ros2/demos](https://github.com/ros2/demos) packages and have the same purpose. I thought we don't need to migrate these packages(`roscpp_tutorials`, `rospy_tutorials`), so we committed to this ros2/demo package. What do you think of this issue?

These packages (`roscpp_packages`, `rospy_tutorials`) shouldn't be ported for the reason you already mentioned - similar packages already exist in ROS 2. When making the PR to the `dashing -devel` branch please just place a `COLCON_IGNORE` file in the directory of these packages and they will be ignores - no need to touch their code.

mlanting:
@routiful and @robotpilot do you mind if I submit a PR for my version of the port on that repository then? You beat me to the PR and I don't want to step on your toes, but I'm just about ready to go with a PR on that repo with all the original functionality working (as far as I can tell with just my own testing so far).

routiful:
@dirk-thomas  
Ok, I have closed this issue. 
I open new PR in ros-tutorials(https://github.com/ros/ros_tutorials/pull/53)

@mlanting 
We already have proposed port this.
(https://github.com/ros2/ros2/issues/607#issuecomment-487521767)
I think yours and ours are functionality working well in ROS2.
If you thought my repo had some issues, please feel free to requst me anytime.

PR #332, opened 2019-05-07
changes to avoid deprecated API's

wjwwood:
Connects to ros2/rclcpp#713

### DIFF
--- a/demo_nodes_cpp/src/topics/talker.cpp
+++ b/demo_nodes_cpp/src/topics/talker.cpp
@@ -56,9 +56,8 @@ public:
       };
 
     // Create a publisher with a custom Quality of Service profile.
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
-    custom_qos_profile.depth = 7;
-    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, custom_qos_profile);
+    rclcpp::QoS qos(rclcpp::KeepLast(7));
+    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, qos);
 
     // Use a timer to schedule periodic message publishing.
     timer_ = this->create_wall_timer(1s, publish_message);
--- a/demo_nodes_cpp/src/topics/talker_serialized_message.cpp
+++ b/demo_nodes_cpp/src/topics/talker_serialized_message.cpp
@@ -113,9 +113,8 @@ public:
         pub_->publish(serialized_msg_);
       };
 
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
-    custom_qos_profile.depth = 7;
-    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, custom_qos_profile);
+    rclcpp::QoS qos(rclcpp::KeepLast(7));
+    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, qos);
 
     // Use a timer to schedule periodic message publishing.
     timer_ = this->create_wall_timer(1s, publish_message);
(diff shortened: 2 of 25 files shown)


image_tools/src/cam2image.cpp
```diff
@@ -140,11 +138,8 @@ int main(int argc, char * argv[])
     };
 
   // Set the QoS profile for the subscription to the flip message.
-  rmw_qos_profile_t custom_flip_qos_profile = rmw_qos_profile_sensor_data;
-  custom_flip_qos_profile.depth = 10;
-
   auto sub = node->create_subscription<std_msgs::msg::Bool>(
-    "flip_image", callback, custom_flip_qos_profile);
+    "flip_image", rclcpp::SensorDataQoS(), callback);
```
ivanpauno: This is more related with the main PR (https://github.com/ros2/rclcpp/pull/713) that here. But if the idea is to encourage users to always specify history depth, shouldn't this looks like:

```cpp
auto sub = node->create_subscription<std_msgs::msg::Bool>(
    "flip_image", rclcpp::SensorDataQoS(5), callback);
```

Or, is fine having some default profiles with an implicit history depth?

demo_nodes_cpp/src/topics/talker.cpp
```diff
@@ -56,9 +56,8 @@ class Talker : public rclcpp::Node
       };
 
     // Create a publisher with a custom Quality of Service profile.
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
-    custom_qos_profile.depth = 7;
-    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, custom_qos_profile);
+    rclcpp::QoS qos(rclcpp::KeepLast(7));
+    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, qos);
```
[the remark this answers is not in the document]
wjwwood: I intentionally left this one like this, because the comment is that we're demonstrating how to create a custom QoS (changing the depth to 7 from the default of 10 is kind of meaningless other than just for demonstration), so I decided in this case only to leave the explicit initialization of the QoS and show how they could create one and customize it before passing it.

demo_nodes_cpp/src/topics/talker_serialized_message.cpp
```diff
@@ -113,9 +113,8 @@ class SerializedMessageTalker : public rclcpp::Node
         pub_->publish(serialized_msg_);
       };
 
-    rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
-    custom_qos_profile.depth = 7;
-    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, custom_qos_profile);
+    rclcpp::QoS qos(rclcpp::KeepLast(7));
+    pub_ = this->create_publisher<std_msgs::msg::String>(topic_name, qos);
```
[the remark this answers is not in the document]
wjwwood: I kept this one for the same reason, just to demonstrate that the qos is created and passed the same way, even for serialized messages.

image_tools/src/cam2image.cpp
wjwwood: Yeah, so this one, is something worth discussing, I'll probably bring it up at the ROS 2 meeting if there's time.

Basically, for profiles that use keep last, requiring them to specify the depth, kind of makes sense, for the same reason we ask people to do this for the normal case, however, if the existing rmw profile uses keep all then it doesn't make sense to require a depth. We could require a `QoSInitialization` which is either keep all or keep last + a depth, but the reason I didn't require that is if a profile were to change from defaulting from keep last to keep all, then the user would be unintentionally overriding this setting always. The effect would be that no rmw profile's history settings would ever be used as the default, making them kind of meaningless.

Maybe we want to go that way, but I'm not sure.

wjwwood: Based on the discussion in our meeting, we think in response to @ivanpauno's question:

> Or, is fine having some default profiles with an implicit history depth?

Yes, especially for cases like `rclcpp::ServicesQoS`, but also for less obvious cases like `rclcpp::SensorDataQoS`.

Therefore, I'll leave this as-is.

PR #350, opened 2019-05-23
Be explicit about reliability in Lifespan demos

emersonknapp:
Explicitly specify Reliability policy in the Lifespan demos to be less dependent on defaults.

The Python version of this demo brought up the issue https://github.com/ros2/rclpy/issues/353, which turned out to just be a difference in default settings between `rclcpp` and `rclpy`. The problem is fixed by making those defaults the same between clients (https://github.com/ros2/rclpy/pull/356) and there is value in that use case, but there is a separate value in being explicit about the settings needed for this functionality to work as expected.

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/quality_of_service_demo/rclpy/quality_of_service_demo_py/lifespan.py
+++ b/quality_of_service_demo/rclpy/quality_of_service_demo_py/lifespan.py
@@ -21,6 +21,7 @@ from rclpy.duration import Duration
 from rclpy.executors import SingleThreadedExecutor
 from rclpy.qos import QoSDurabilityPolicy
 from rclpy.qos import QoSProfile
+from rclpy.qos import QoSReliabilityPolicy
 
 
 def parse_args():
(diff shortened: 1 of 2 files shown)


quality_of_service_demo/rclpy/quality_of_service_demo_py/lifespan.py
```diff
@@ -51,7 +52,10 @@ def main(args=None):
 
     qos_profile = QoSProfile(
         depth=parsed_args.history,
```
jacobperron: Unrelated to this PR: I would recommend making arguments match the QoS setting for clarity. Briefly, this looked like a bug, assigning a history setting to a depth setting.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7215/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3413/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5909/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7039/)

PR #361, opened 2019-05-30
Fix memory leak in the dummy robot examples.

### LINKED ISSUE #340: got a memory leak
these code in files like  dummy_robot/**/*.cpp :
rclcpp::TimeSource ts(node);
rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
ts.attachClock(clock);
in the while loop cause the memory leak, they should be put out of the where loop, and then, I tried it ok.

clalancette:
They were repeatedly creating a new clock in the loop, and then
leaking that clock.  Instead, create the clock once at the top
of the loop.

This should fix #340

### DIFF
--- a/dummy_robot/dummy_map_server/src/dummy_map_server.cpp
+++ b/dummy_robot/dummy_map_server/src/dummy_map_server.cpp
@@ -53,6 +53,10 @@ int main(int argc, char * argv[])
     msg.data.push_back(-1);
   }
 
+  rclcpp::TimeSource ts(node);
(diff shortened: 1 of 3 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7265/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3445/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5945/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7081/)

Issue #325, opened 2019-04-24
Update demos to use composition


jacobperron:
#### Feature description

Since it is the recommended way to write nodes as components, it would be good to refactor most, if not all, of the demos to be composable. This way new users learning from the demos will be exposed to best practices in ROS 2 development.

Since components are only available in C++, there is nothing to be done for the Python demos.

#### Implementation considerations

The [composition](https://github.com/ros2/demos/tree/0cb8fc5ff993820d33826d2de23fe81a3c8a4062/composition) package is a good place to look for examples of writing components.

nuclearsandwich:
Can this be closed now that #375 is merged/

PR #367, opened 2019-06-05
Refactored image_tools demo. Nodes written as components. 

skucheria:
Addresses https://github.com/ros2/demos/issues/325


image_tools/CMakeLists.txt
```diff
@@ -13,37 +13,63 @@ endif()
...
+  src/cam2image_main.cpp
+  src/options.cpp
+  src/burger.cpp)
+target_link_libraries(cam2image
+  showimage_component
+  cam2image_component)
+ament_target_dependencies(cam2image
+  "rclcpp")
+
+  add_executable(showimage
```
jacobperron: nit: unindent

image_tools/CMakeLists.txt
```diff
@@ -13,37 +13,63 @@ endif()
...
+  showimage_component
+  cam2image_component
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin)
 
 install(TARGETS
-  cam2image
   showimage
+  cam2image
```
jacobperron: nit: This change is unnecessary (I guess they were purposefully alphabetized before).

image_tools/include/image_tools/showimage.hpp
```diff
@@ -0,0 +1,75 @@
...
+#ifndef IMAGE_TOOLS__SHOW_CONFIG_COMPONENT_HPP_
+#define IMAGE_TOOLS__SHOW_CONFIG_COMPONENT_HPP_
+
+#include <cstdio>
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "opencv2/highgui/highgui.hpp"
```
jacobperron: It doesn't appear that theses OpenCV headers are needed here. Best move them to the `cpp` file where they are needed. Same goes for any other headers that are only needed in the `cpp` file.

image_tools/include/image_tools/showimage.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+	IMAGE_TOOLS_PUBLIC
+	bool setup(int argc, char **argv);
+
+	IMAGE_TOOLS_PUBLIC
+	void execute();
+
+private: 
+	rclcpp::Subscription<sensor_msgs::msg::Image>::SharedPtr sub_;
+	int numArgs;
```
jacobperron: Use `snake_case` and be consistent with trailing underscores for class members.

image_tools/src/showimage_main.cpp
```diff
@@ -0,0 +1,52 @@
...
+  rclcpp::executors::SingleThreadedExecutor exec;
+  rclcpp::NodeOptions options;
+
+  // Add some nodes to the executor which provide work for the executor during its "spin" function.
+  // An example of available work is executing a subscription callback, or a timer callback.
+
+  auto showimage = std::make_shared<image_tools::ShowImage>(options);
+  if (showimage->setup(argc, argv)){
+    exec.add_node(showimage);
+    showimage->execute();
```
jacobperron: Currently, this doesn't work if I want to load into a container dynamically, e.g.

```sh
ros2 component standalone image_tools image_tools::ShowImage
```

What if we called the execute method in the constructor (or as a callback of a oneshot timer started in the constructor)? Rather than having a setup method, we could overload the constructor for the purpose of keeping the demo backwards compatible, i.e.

```c++
ShowImage(rclcpp::NodeOptions options);
ShowImage(rclcpp::NodeOptions options, int argc, char ** argv);
```

Both constructors call the execute method, and the latter one does the additional setup.  The first constructor could also make use of ROS parameters for the setup (this can be done in a follow-up PR if you like).

I'm not sure what the best practice is for when there is an error initializing a component. In this case we could throw a runtime error from the constructor, catch it in `main` and exit early.

image_tools/src/showimage.cpp
```diff
@@ -80,73 +82,70 @@ void show_image(
...
-  if (show_camera) {
+/// Execute main functions with image subscriber and callback 
+/**
+ */
+void ShowImage::execute(){
+if (show_camera) {
     // Initialize an OpenCV named window called "showimage".
     cv::namedWindow("showimage", cv::WINDOW_AUTOSIZE);
-    cv::waitKey(1);
+    cv::waitKey(10);
```
jacobperron: Why change to `10`?

image_tools/src/showimage.cpp
```diff
@@ -80,73 +82,70 @@ void show_image(
   }
 }
 
-int main(int argc, char * argv[])
+/// Constructor that initializes the default demo parameters
+/**
+ */
+ShowImage::ShowImage(rclcpp::NodeOptions options)
```
jacobperron: The options parameter can be `const &`

image_tools/src/showimage.cpp
```diff
@@ -80,73 +82,70 @@ void show_image(
...
-  setvbuf(stdout, NULL, _IONBF, BUFSIZ);
-
-  // Configure demo parameters with command line options.
-  if (!parse_command_options(
-      argc, argv, &depth, &reliability_policy, &history_policy, &show_camera, nullptr, nullptr,
-      nullptr, nullptr, &topic))
-  {
-    return 0;
-  }
+  depth = rmw_qos_profile_default.depth;
```
jacobperron: These can be set in the constructor's initializer list.

image_tools/CMakeLists.txt
```diff
@@ -13,37 +13,63 @@ endif()
...
+ament_target_dependencies(cam2image_component
   "rclcpp"
   "sensor_msgs"
   "std_msgs"
-  "OpenCV")
+  "OpenCV"
+  "rclcpp_components")
+rclcpp_components_register_nodes(cam2image_component "image_tools::Cam2Image")
+
+add_executable(cam2image
```
dirk-thomas: That should happen as part of the components CMake API - maybe optionally controlled by a flag - but the user code shouldn't need to do it explicitly.

image_tools/CMakeLists.txt
```diff
@@ -29,7 +29,6 @@ ament_target_dependencies(showimage_component
   "std_msgs"    
   "rclcpp_components"
   "OpenCV")
-rclcpp_components_register_nodes(showimage_component "image_tools::ShowImage")
```
jacobperron: Where to the components get registered otherwise?

jacobperron: This should be added back, otherwise the components are not registered with the resource index, i.e. added to the file `install/image_tools/share/ament_index/resource_index/rclcpp_components/image_tools`.

image_tools/CMakeLists.txt
jacobperron: Correct me if I'm wrong, but this looks like something that is not implemented yet in the components CMake API. I believe this is the relevant ticket: https://github.com/ros2/rclcpp/issues/114

@skucheria I would add a TODO here (and above the other `add_executable`) stating that the components API should handle this for us.

image_tools/CMakeLists.txt
skucheria: Ok, just pushed changes

image_tools/CMakeLists.txt
dirk-thomas: I think the referenced feature needs to land before this can be merged.

image_tools/src/cam2image.cpp
```diff
@@ -71,92 +67,80 @@ void convert_frame_to_message(
...
 
     // Set the width and height based on command line arguments.
-    cap.set(cv::CAP_PROP_FRAME_WIDTH, static_cast<double>(width));
-    cap.set(cv::CAP_PROP_FRAME_HEIGHT, static_cast<double>(height));
+    cap.set(cv::CAP_PROP_FRAME_WIDTH, static_cast<double>(width_));
+    cap.set(cv::CAP_PROP_FRAME_HEIGHT, static_cast<double>(height_));
     if (!cap.isOpened()) {
       RCLCPP_ERROR(node_logger, "Could not open video stream");
-      return 1;
+      // return 1;
```
jacobperron: It would be good if the demo stops running when there is an error (same as the previous behavior). I think we can achieve this by throwing an exception here:

```c++
throw std::runtime_error("Could not open video stream");
```

I would still keep the log macro since it looks like the component API does not echo the exception message.

image_tools/src/showimage.cpp
```diff
@@ -17,22 +17,28 @@
...
-
 #include "rclcpp/rclcpp.hpp"
 
 #include "sensor_msgs/msg/image.hpp"
 
 #include "image_tools/options.hpp"
 
+#include "image_tools/showimage.hpp"
+
+#include "opencv2/highgui/highgui.hpp"
```
jacobperron: style nit: move opencv header includes where they were (above rclcpp/rclcpp.hpp). See [style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) for order of includes.

image_tools/src/showimage.cpp
```diff
@@ -80,73 +86,75 @@ void show_image(
   }
 }
 
-int main(int argc, char * argv[])
+/// Default constructor for no command line args
+/**
+ */
```
jacobperron: Remove empty doc block (or populate). 

Also, I think all of the doc blocks can be moved to the header files.

image_tools/src/showimage.cpp
```diff
@@ -80,73 +86,75 @@ void show_image(
...
+}
 
-  if (show_camera) {
+/// Execute main functions with image subscriber and callback
+/**
+ */
+void ShowImage::execute(){
+  if (show_camera_) {
     // Initialize an OpenCV named window called "showimage".
+      // RCLCPP_INFO(this->get_logger(), "SHOWING THE CAMERA");
```
jacobperron: Remove.

image_tools/src/showimage.cpp
```diff
@@ -80,73 +86,75 @@ void show_image(
...
     cv::waitKey(1);
   }
 
-  // Initialize a ROS node.
-  auto node = rclcpp::Node::make_shared("showimage");
-
-  // Set quality of service profile based on command line options.
   auto qos = rclcpp::QoS(
     rclcpp::QoSInitialization(
-      // The history policy determines how messages are saved until taken by
```
jacobperron: I think it's good to leave these comments in since this is code that we expect users to look at.

image_tools/src/showimage.cpp
```diff
@@ -80,73 +86,75 @@ void show_image(
...
-      // specified by the "depth" parameter.
-      history_policy,
-      // Depth represents how many messages to store in history when the
-      // history policy is KEEP_LAST.
-      depth
+      history_policy_,
+      depth_
   ));
-
-  // The reliability policy can be reliable, meaning that the underlying transport layer will try
```
jacobperron: Add this comment back as well.

image_tools/src/showimage.cpp
```diff
@@ -80,73 +86,75 @@ void show_image(
...
-      show_image(msg, show_camera, node->get_logger());
+      show_image(msg, show_camera_, this->get_logger());
     };
 
-  std::cerr << "Subscribing to topic '" << topic << "'" << std::endl;
-  RCLCPP_INFO(node->get_logger(), "Subscribing to topic '%s'", topic.c_str());
-  // Initialize a subscriber that will receive the ROS Image message to be displayed.
-  auto sub = node->create_subscription<sensor_msgs::msg::Image>(
-    topic, qos, callback);
+  std::cerr << "Subscribing to topic '" << topic_ << "'" << std::endl;
```
jacobperron: I think this can be removed. It looks like debugging leftovers and there is already a log macro with the same info.

image_tools/src/showimage.cpp
```diff
@@ -80,73 +86,75 @@ void show_image(
...
 
-  rclcpp::shutdown();
+/// Read in and parse command line arguments.
+/**
+ * \param[in] argc 
+ * \param[in] argv 
+ * \return A bool whether command line options were valid or not 
+ */
+bool ShowImage::setup(int argc, char ** argv){
+  if (!parse_command_options(
```
jacobperron: Just return the result of `parse_command_options`.

image_tools/CMakeLists.txt
```diff
@@ -11,40 +11,50 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
...
-
+find_package(rcutils REQUIRED)
 find_package(OpenCV REQUIRED)
 
 include_directories(include)
+include(${ADD_NODE_PATH})
 
-add_executable(cam2image
-  src/burger.cpp
+rclcpp_components_add_node(TARGET showimage_component SRCS
```
jacobperron: With ros2/rclcpp#758, we can rename our nodes back to their original names (e.g. `showimage` instead of `showimage_component`).

image_tools/src/cam2image.cpp
```diff
@@ -71,92 +67,91 @@ void convert_frame_to_message(
...
-  // Create the image publisher with our custom QoS profile.
-  auto pub = node->create_publisher<sensor_msgs::msg::Image>(topic, qos);
+  qos.reliability(reliability_policy_);
+  RCLCPP_INFO(node_logger, "Publishing data on topic '%s'", topic_.c_str());
+  pub_ = create_publisher<sensor_msgs::msg::Image>(topic_, qos);
 
   // is_flipped will cause the incoming camera image message to flip about the y-axis.
   bool is_flipped = false;
-
-  // Subscribe to a message that will toggle flipping or not flipping, and manage the state in a
```
jacobperron: nit: I'd leave this comment in since it is demo code we expect new users to look at.

image_tools/src/cam2image.cpp
```diff
@@ -71,92 +67,91 @@ void convert_frame_to_message(
...
-  // callback.
-  auto callback =
-    [&is_flipped, &node_logger](const std_msgs::msg::Bool::SharedPtr msg) -> void
+  auto callback = [&is_flipped, &node_logger](const std_msgs::msg::Bool::SharedPtr msg) -> void
     {
       is_flipped = msg->data;
       RCLCPP_INFO(node_logger, "Set flip mode to: %s", is_flipped ? "on" : "off");
     };
 
-  // Set the QoS profile for the subscription to the flip message.
```
jacobperron: nit: Same; keep the comment.

image_tools/src/cam2image.cpp
```diff
@@ -165,15 +160,13 @@ int main(int argc, char * argv[])
   cv::Mat flipped_frame;
 
   size_t i = 1;
-
-  // Our main event loop will spin until the user presses CTRL-C to exit.
```
jacobperron: nit: keep comment.

image_tools/src/cam2image_main.cpp
```diff
@@ -0,0 +1,45 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
```
jacobperron: We can remove this file now that rclcpp_component is handling building the executable.

image_tools/src/showimage_main.cpp
```diff
@@ -0,0 +1,47 @@
+// Copyright 2016 Open Source Robotics Foundation, Inc.
```
jacobperron: Remove this file.

image_tools/src/showimage.cpp
```diff
@@ -80,40 +85,34 @@ void show_image(
...
-  {
-    return 0;
+  execute();
+}
+
+/// Constructor for setup with command line args
+ShowImage::ShowImage(const rclcpp::NodeOptions & options, int argc, char ** argv)
+: Node("showimage", options)
+{
+  if (setup(argc, argv)) {
```
jacobperron: Now that we've moved the responsibility of building the executable to rclcpp_component, I think we should revisit grabbing the arguments from the `NodeOptions` instead.

image_tools/src/showimage.cpp
```diff
@@ -122,31 +122,42 @@ int main(int argc, char * argv[])
...
+ * \return A bool whether command line options were valid or not
+ */
+bool ShowImage::setup(int argc, const char ** argv)
+{
+  return parse_command_options(
+    argc, argv, &depth_, &reliability_policy_, &history_policy_, &show_camera_, nullptr, nullptr,
+    nullptr, nullptr, &topic_);
 }
+
+
```
jacobperron: nit: remove extra vertical whitespace. Only one empty line is necessary (same below).

image_tools/src/showimage.cpp
```diff
@@ -80,40 +86,34 @@ void show_image(
...
-  if (show_camera) {
+  if (setup(argc, argv)) {
+    execute();
+  } else {
+    rclcpp::shutdown();
+  }
+}
+
+/// Execute main functions with image subscriber and callback
+/**
```
jacobperron: nit: remove empty comment.
Also, any doc-blocks should go in the header.

image_tools/src/cam2image.cpp
```diff
@@ -71,92 +67,93 @@ void convert_frame_to_message(
...
+ * \return A bool whether command line options were valid or not
+ */
+bool Cam2Image::setup(int argc, const char ** argv)
+{
+  return parse_command_options(
+    argc, argv, &depth_, &reliability_policy_, &history_policy_, &show_camera_, &freq_, &width_,
+    &height_, &burger_mode_, &topic_);
+}
+
+/// Execute main functions of component
```
jacobperron: Move all function/class documentation to the header files.

image_tools/src/showimage.cpp
```diff
@@ -80,40 +86,34 @@ void show_image(
...
-
-  // Configure demo parameters with command line options.
-  if (!parse_command_options(
-      argc, argv, &depth, &reliability_policy, &history_policy, &show_camera, nullptr, nullptr,
-      nullptr, nullptr, &topic))
-  {
-    return 0;
+  std::vector<std::__cxx11::basic_string<char>> args = options.arguments();
+  int argc = options.arguments().size();
+  const char **argv = new const char*[args.size()];
```
jacobperron: Converting to `char**` seems unnecessary. Since `setup()` is converting back to a `std::vector<std::string>` anyways, I would just change the signature of the function to accept a vector of strings.

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7377/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3535/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6052/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7205/)

image_tools/package.xml
```diff
@@ -12,11 +12,15 @@
   <build_depend>rclcpp</build_depend>
   <build_depend>sensor_msgs</build_depend>
   <build_depend>std_msgs</build_depend>
+  <build_depend>rclcpp</build_depend>
```
jacobperron: This is listed as a build depend twice (previously at L12).

image_tools/package.xml
```diff
@@ -12,11 +12,15 @@
...
+  <build_depend>rclcpp_components</build_depend>
```
jacobperron: Put this in alphabetical order.

image_tools/package.xml
```diff
@@ -12,11 +12,15 @@
...
 
   <exec_depend>libopencv-dev</exec_depend>
   <exec_depend>rclcpp</exec_depend>
   <exec_depend>sensor_msgs</exec_depend>
   <exec_depend>std_msgs</exec_depend>
+  <exec_depend>rclcpp</exec_depend>
+  <exec_depend>rclcpp_components</exec_depend>
```
jacobperron: Alphabetize

image_tools/package.xml
```diff
@@ -12,11 +12,15 @@
...
   <build_depend>sensor_msgs</build_depend>
   <build_depend>std_msgs</build_depend>
+  <build_depend>rclcpp</build_depend>
+  <build_depend>rclcpp_components</build_depend>
 
   <exec_depend>libopencv-dev</exec_depend>
   <exec_depend>rclcpp</exec_depend>
   <exec_depend>sensor_msgs</exec_depend>
   <exec_depend>std_msgs</exec_depend>
+  <exec_depend>rclcpp</exec_depend>
```
jacobperron: Already listed as an exec_depend.

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7513/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3649/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6153/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7338/)

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7514/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3650/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6154/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7339/)

PR #385, opened 2019-08-09
Pendulum fixes

### LINKED ISSUE #386: test_pendulum failing with nonsensical statistics on linux-aarch64
## Bug report

**Required Info:**

- Operating System:
  - Jenkins Linux Arm64
- Installation type:
  - From Source
- Version or commit hash:
  - 9efc94fce1123870c9bad33c966a5d81c977826a
- DDS implementation:
  - OpenSplice
  - N/A

#### Steps to reproduce issue

Run CI with OpenSplice

#### Expected behavior

Test "test_pendulum__rmw_opensplice_cpp" passes.

#### Actual behavior

Test fails and reports absurd statistics. Current latency is very high (1.7 seconds), mean latency is absurdly high (uninitialized memory?). Min latency is higher than max latency. Number of pagefaults is also absurd.

> [INFO] [pendulum_demo-2]: process started with pid [55617]
> [pendulum_demo-2] Couldn't set scheduling priority and policy: Operation not permitted
> [pendulum_demo-2] mlockall failed: Cannot allocate memory
> [pendulum_demo-2] Couldn't lock all cached virtual memory.
> [pendulum_demo-2] Pagefaults from reading pages not yet mapped into RAM will be recorded.
> [pendulum_logger-1] Logger node initialized.
> [pendulum_logger-1] Commanded motor angle: 1.570796
> [pendulum_logger-1] Actual motor angle: 1.561280
> [pendulum_logger-1] Current latency: 1699557180 ns
> [pendulum_logger-1] Mean latency: 4458224516038787409289673575093816669551304146650037671236339690561781237058564279746291212376660112936812899278369000229134709514347391053183640763362416971774278560649182152566312699581486525144027387385646723574136033038253156096752904762288198647808.000000 ns
> [pendulum_logger-1] Min latency: 1835355439 ns
> [pendulum_logger-1] Max latency: 1047684450 ns
> [pendulum_logger-1] Minor pagefaults during execution: 8459008377404276325
> [pendulum_logger-1] Major pagefaults during execution: 7237087991477200236
> [pendulum_logger-1] 

https://ci.ros2.org/job/ci_linux-aarch64/3926/testReport/junit/projectroot/test_pendulum__rmw_opensplice_cpp/test_pendulum__rmw_opensplice_cpp/

clalancette:
This PR, along with https://github.com/ros2/realtime_support/pull/81 , should fix an issue in the `pendulum_control` demo where we publish uninitialized data.  There are three fixes in here, corresponding to the 3 commits:

1.  Remove some unused code in `rtt_executor.hpp`.  This isn't strictly required here, but there is no reason for this code to be here since it is never used.
1.  Switch the min_latency and max_latency fields in the RttestResults.msg to be signed integers.  The underlying latency data coming from `realtime_tools` is signed, so this should be signed as well.
1.  Make it so that we don't publish data if it hasn't yet been initialized.

With this in place, it looks to me like the uninitialized data problems that we have with pendulum_control tests should be gone.  This *may* solve https://github.com/ros2/build_cop/issues/133 ; I'll see what happens when CI comes back.  @cottsay FYI as current build farmer.

Fixes https://github.com/ros2/build_cop/issues/133
Fixes #386

### DIFF
--- a/pendulum_control/include/pendulum_control/rtt_executor.hpp
+++ b/pendulum_control/include/pendulum_control/rtt_executor.hpp
@@ -59,8 +59,11 @@ public:
     return rclcpp::ok() && running;
   }
 
-  void set_rtt_results_message(pendulum_msgs::msg::RttestResults & msg) const
+  bool set_rtt_results_message(pendulum_msgs::msg::RttestResults & msg) const
   {
+    if (!results_available) {
+      return false;
+    }
     msg.cur_latency = last_sample;
     msg.mean_latency = results.mean_latency;
     msg.min_latency = results.min_latency;
@@ -71,6 +74,8 @@ public:
     clock_gettime(CLOCK_MONOTONIC, &curtime);
     msg.stamp.sec = curtime.tv_sec;
     msg.stamp.nanosec = curtime.tv_nsec;
+
+    return true;
   }
 
   /// Wrap executor::spin into rttest_spin.
@@ -105,7 +110,9 @@ public:
     executor->spin_some();
 
     // Retrieve rttest statistics accumulated so far and store them in the executor.
-    rttest_get_statistics(&executor->results);
+    if (rttest_get_statistics(&executor->results) >= 0) {
+      executor->results_available = true;
+    }
     rttest_get_sample_at(executor->results.iteration, &executor->last_sample);
     // In case this boolean wasn't set, notify that we've recently run the callback.
     executor->running = true;
@@ -114,6 +121,7 @@ public:
 
   // For storing accumulated performance statistics.
   rttest_results results;
+  bool results_available{false};
   // True if the executor is spinning.
   bool running;
   // True if rttest has initialized and hasn't been stopped yet.
--- a/pendulum_control/include/pendulum_control/rtt_executor.hpp
+++ b/pendulum_control/include/pendulum_control/rtt_executor.hpp
@@ -52,25 +52,11 @@ public:
   /// Default destructor
   virtual ~RttExecutor() {}
 
-  /// Return the status of rttest.
-  // \return True if rttest has initialized, false if it is uninitialized or has finished.
-  bool is_rttest_ready() const
-  {
-    return rttest_ready;
-  }
-
   /// Return true if the executor is currently spinning.
   // \return True if rclcpp is running and if the "running" boolean is set to true.
   bool is_running() const
   {
     return rclcpp::ok() && running;
-  }
-
-  /// Retrieve the results measured by rttest
-  // \param[in] output A struct containing performance statistics.
-  void get_rtt_results(rttest_results & output) const
-  {
-    output = results;
   }
 
   void set_rtt_results_message(pendulum_msgs::msg::RttestResults & msg) const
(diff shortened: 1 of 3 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7857/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3920/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6411/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7717/)

clalancette:
Given that this was failing on aarch64 every night for quite a while with a very similar configuration, I'd say this actually fixes it.  Woohoo!  I'll update the description.

clalancette:
I had a silly bug in here, so here is another CI run:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7877/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3933/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6425/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7734/)

clalancette:
And there was one more bug.  Now I think I really have it fixed.  Another CI to see if I'm correct:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7881/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3936/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6430/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7749/)

dirk-thomas:
> Remove some unused code in `rtt_executor.hpp`.  This isn't strictly required here, but there is no reason for this code to be here since it is never used.

It would be good to keep changes like this in a separate PR. Simply because it makes reviewing each PR easier and each one can iterate and get merged independently. (Doesn't have to be for this change now but please do that in the future.)

> Switch the min_latency and max_latency fields in the RttestResults.msg to be signed integers. The underlying latency data coming from `realtime_tools` is signed, so this should be signed as well.

How can the values be signed? If they conceptionally can't I would advocate to keep the unsigned in the message and do the necessary conversion (or consider updating `realtime_tools`).

> Make it so that we don't publish data if it hasn't yet been initialized.

:+1:

clalancette:
I finally figured this out; it was an overflow.  This now requires https://github.com/ros2/realtime_support/pull/82 , and should now fix the pendulum failures for real.  Here's some more CI for it:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8190/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4168/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6670/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8066/)

pendulum_control/include/pendulum_control/rtt_executor.hpp
```diff
@@ -52,31 +52,22 @@ class RttExecutor : public rclcpp::executor::Executor
...
-
-  void set_rtt_results_message(pendulum_msgs::msg::RttestResults & msg) const
+  bool set_rtt_results_message(pendulum_msgs::msg::RttestResults & msg) const
   {
+    if (!results_available) {
+      return false;
+    }
     msg.cur_latency = last_sample;
     msg.mean_latency = results.mean_latency;
+    assert(results.min_latency >= 0);
```
jacobperron: I guess it's not hurting to have these assertions here, but it seems superfluous since `min_latency` and `max_latency` are non-negative by declaration:

https://github.com/ros2/demos/blob/97651912ebad439a6bc62c34671cad6a5bfee785/pendulum_msgs/msg/RttestResults.msg#L8-L9

clalancette: It's a little complicated, but the underlying code in https://github.com/ros2/realtime_support/blob/aef9318be62e1b63753d55daab9606510f0b9c46/rttest/src/rttest.cpp actually allows for negative numbers.  The idea with the assertions here is to make it obvious (via a test failure) if we ever try to stuff a negative number into the unsigned 64-bit quantity.  If we remove these, then it could lead to a difficult-to-figure bug where we see a giant number in the unsigned field.  It's a somewhat arbitrary distinction (it's a bug either way), but this just makes it a bit more obvious.  Does that make sense?

jacobperron: Nevermind, I my brain parsed this as asserting on the message type. I see now `results` has type `rttest_results`. This makes sense.

PR #394, opened 2019-09-29
zero copy api

Karsten1987:
connects to https://github.com/ros2/ros2/issues/785
Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/demo_nodes_cpp/src/topics/talker_loaned_message.cpp
+++ b/demo_nodes_cpp/src/topics/talker_loaned_message.cpp
@@ -15,9 +15,7 @@
 #include <chrono>
 #include <cstdio>
 #include <memory>
-#include <string>
 #include <utility>
-#include <vector>
 
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp_components/register_node_macro.hpp"
@@ -87,8 +85,8 @@ public:
 
     // Create a publisher with a custom Quality of Service profile.
     rclcpp::QoS qos(rclcpp::KeepLast(7));
-    pod_pub_ = this->create_publisher<std_msgs::msg::Float64>("/chatter_pod", qos);
-    non_pod_pub_ = this->create_publisher<std_msgs::msg::String>("/chatter", qos);
+    pod_pub_ = this->create_publisher<std_msgs::msg::Float64>("chatter_pod", qos);
+    non_pod_pub_ = this->create_publisher<std_msgs::msg::String>("chatter", qos);
 
     // Use a timer to schedule periodic message publishing.
     timer_ = this->create_wall_timer(1s, publish_message);
@@ -99,7 +97,6 @@ private:
   rclcpp::Publisher<std_msgs::msg::Float64>::SharedPtr pod_pub_;
   rclcpp::Publisher<std_msgs::msg::String>::SharedPtr non_pod_pub_;
   rclcpp::TimerBase::SharedPtr timer_;
-  std::string topic_name_ = "chatter";
 };
 
 }  // namespace demo_nodes_cpp
--- a/demo_nodes_cpp/src/topics/talker_loaned_message.cpp
+++ b/demo_nodes_cpp/src/topics/talker_loaned_message.cpp
@@ -54,10 +54,20 @@ public:
       auto publish_message =
         [this]() -> void
         {
+          // We loan a message here and don't allocate the memory on the stack.
+          // For middlewares which support message loaning, this means the middleware
+          // completely owns the memory for this message.
+          // This enables a zero-copy message transport for middlewares with shared memory
+          // capabilities.
+          // If the middleware doesn't support this, the loaned message will be allocated
+          // with the allocator instance provided by the publisher.
           auto loaned_msg = pub_->loan_message();
           auto msg_data = "Hello World: " + std::to_string(count_++);
           loaned_msg.get().data = msg_data;
           RCLCPP_INFO(this->get_logger(), "Publishing: '%s'", msg_data.c_str());
+          // As the middleware might own the memory allocated for this message,
+          // a call to publish explicitly transfers ownership back to the middleware.
+          // The loaned message instance is thus no longer valid after a call to publish.
           pub_->publish(std::move(loaned_msg));
         };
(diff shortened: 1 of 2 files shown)


dirk-thomas:
How does the loaning apply to a dynamic size message like `String`?

Karsten1987:
We should highlight this a little bit stronger in the design doc, but essentially it's the middleware's task to declare whether or not it can loan messages of a certain message type.

If yes (for example for fixed sized arrays or POD), the memory for the loaned message is allocated by the middleware. If not, the loaned message is using an alternatively provided instance of an allocator. The idea is that if the middleware doesn't support message loaning, the user should be able to control the allocation.

The logic is implemented here: https://github.com/ros2/rclcpp/pull/864/files#diff-8c68c968941e2e05ba8974455ea3f1b5R73-R74

Independent of the middleware, the user code should always work (whether it is zero-copy or not).

demo_nodes_cpp/src/topics/talker_loaned_message.cpp
```diff
@@ -0,0 +1,117 @@
...
+        [this]() -> void
+        {
+          // We loan a message here and don't allocate the memory on the stack.
+          // For middlewares which support message loaning, this means the middleware
+          // completely owns the memory for this message.
+          // This enables a zero-copy message transport for middlewares with shared memory
+          // capabilities.
+          // If the middleware doesn't support this, the loaned message will be allocated
+          // with the allocator instance provided by the publisher.
+          auto loaned_msg = pub_->loan_message();
```
mjcarroll: Since we are targeting this iteration to POD, does it make sense to use a message that is POD (rather than String).

Karsten1987: I think it's important to show that non-pod still work transparently. Proposal: We introduce two topics, a `pod_chatter` and a `non_pod_chatter` or something similar.

mjcarroll: I'm good with that proposal.

demo_nodes_cpp/src/topics/talker_loaned_message.cpp
```diff
@@ -0,0 +1,143 @@
...
+class LoanedMessageTalker : public rclcpp::Node
+{
+public:
+  DEMO_NODES_CPP_PUBLIC
+  explicit LoanedMessageTalker(const rclcpp::NodeOptions & options)
+  : Node("loaned_message_talker", options)
+  {
+    // Create a function for when messages are to be sent.
+    setvbuf(stdout, NULL, _IONBF, BUFSIZ);
+    std::vector<std::string> args = options.arguments();
```
jacobperron: Please remove logic related to argument parsing. It doesn't add any value to the demo since the topic can be changed via a remap. I am doing the same for the rest of the demos in https://github.com/ros2/demos/pull/401.

Karsten1987: done in https://github.com/ros2/demos/pull/394/commits/5f7dc32ceec8ff02805c82fa18ae719c627f57e3

demo_nodes_cpp/src/topics/talker_loaned_message.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // Use a timer to schedule periodic message publishing.
+    timer_ = this->create_wall_timer(1s, publish_message);
+  }
+
+private:
+  size_t count_ = 1;
+  rclcpp::Publisher<std_msgs::msg::Float64>::SharedPtr pod_pub_;
+  rclcpp::Publisher<std_msgs::msg::String>::SharedPtr non_pod_pub_;
+  rclcpp::TimerBase::SharedPtr timer_;
+  std::string topic_name_ = "chatter";
```
jacobperron: This member is unused and can be removed

demo_nodes_cpp/src/topics/talker_loaned_message.cpp
```diff
@@ -0,0 +1,107 @@
...
+        auto non_pod_loaned_msg = non_pod_pub_->loan_message();
+        auto non_pod_msg_data = "Hello World: " + std::to_string(count_++);
+        non_pod_loaned_msg.get().data = non_pod_msg_data;
+        RCLCPP_INFO(this->get_logger(), "Publishing: '%s'", non_pod_msg_data.c_str());
+        non_pod_pub_->publish(std::move(non_pod_loaned_msg));
+      };
+
+    // Create a publisher with a custom Quality of Service profile.
+    rclcpp::QoS qos(rclcpp::KeepLast(7));
+    pod_pub_ = this->create_publisher<std_msgs::msg::Float64>("/chatter_pod", qos);
```
jacobperron: Is it correct to make this a globally namespaced topic? Then if the node is namespaced, the topic will not change. Just checking if we want this behavior or if this should be changed to `"chatter_pod"`. Same for the topic.

demo_nodes_cpp/src/topics/talker_loaned_message.cpp
```diff
@@ -0,0 +1,107 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <chrono>
+#include <cstdio>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
```
jacobperron: This header is no longer needed.

demo_nodes_cpp/src/topics/talker_loaned_message.cpp
```diff
@@ -0,0 +1,107 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <chrono>
+#include <cstdio>
+#include <memory>
+#include <string>
```
jacobperron: I don't believe this header is needed anymore.

PR #396, opened 2019-10-11
Prevent argparse from parsing ROS args in Python demos

hidmic:
Precisely what the title says. If ROS args are not removed, `argparse` chokes on them.

Depends on https://github.com/ros2/rcl/pull/518.


hidmic:
CI for `demo_nodes_py` (linters only):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8373/)

hidmic:
Alright, moving forward.

Issue #390, opened 2019-09-14
Replace custom CLI arguments with ROS parameters


jacobperron:
#### Feature description

By switching to ROS parameters, it should hopefully cut down on LOC and set a better example for users.

For example, [image_tools/src/options.cpp](https://github.com/ros2/demos/blob/master/image_tools/src/options.cpp) can be removed entirely once we switch to ROS params.

jacobperron:
Blocked by https://github.com/ros2/rclcpp/issues/860

jacobperron:
CLI arguments were changed to ROS parameters in #398 and removed altogether in #401.

The last C++ demo code using CLI arguments are the QoS demos, but they would need some heavy refactoring if we want to switch them to ROS parameters. Currently they parse arguments for multiple nodes per executable. I believe it should be possible for each node to only check for the parameters it cares about inside their constructors.

There are some Python nodes that make use of argparse to read CLI arguments, and a case has been made that we lose the benefit of completion if we stop using it (https://github.com/ros2/demos/pull/398#issuecomment-542417171). But, it looks to me that we are not taking advantage of completion in argparse (ie. it's not enabled) in any of our demos, so maybe it's okay that we use ROS parameters instead. On the other hand it is pretty convenient from an implementation and user standpoint (help formatting and CLI options look nice), so I'm inclined to leave the demos using argparse as-is.

jacobperron:
I don't think there's anything left to do for this ticket.

PR #398, opened 2019-10-15
[image_tools] Use ROS parameters instead of regular CLI arguments

jacobperron:
Using of ROS parameters make the code a little tidier, though a little more verbose on the command-line.
IMO, it seems like a good way to demonstrate best practices as well.

Example usage:

```sh
$ ros2 run image_tools cam2image --ros-args -p burger_mode:=true -p reliability:=reliable
```
```sh
$ ros2 run image_tools showimage --ros-args -p reliability:=reliable
```

Connects to #390


dirk-thomas:
How does a user get a similar help than what was provided by `--help` before?

jacobperron:
> How does a user get a similar help than what was provided by --help before?

Good question, I guess there's no equivalent to `--help`.

Maybe this is a good example for a feature/tool that is lacking for ROS parameters.
At the moment, you can list parameters for the node, but only after it is running:

```sh
$ ros2 param list
/showimage:
  depth
  history
  reliability
  show_image
  use_sim_time
```

In addition it would be nice if we could see the types/constraints on these params (https://github.com/ros2/ros2cli/issues/362).

And even better (equivalent to `--help`) would be to introspect the node before actually running it. I imagine some like

```sh
$ ros2 run image_tools showimage --ros-args --help
Parameters:
    <list of parameters and descriptors>
Publishers:
    <list of publisher types and topics>
Subscriptions:
    <list of subscription types and topics>
... etc for services and actions.
```

I suppose this kind of introspection feature needs a lot more thought and discussion to make it happen.

dirk-thomas:
Not relevant in this case where the node is implemented in C++. But in the case of Python nodes using `argparse` another major advantage is getting free `argcomplete` functionality. So switching a Python node to parameters would loose that too which I think would be undesired.

jacobperron:
I've added back an option for help (`-h` / `--help`): da736d0

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8415/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4326/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6850/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8314/)

image_tools/src/cam2image.cpp
```diff
@@ -118,55 +96,151 @@ class Cam2Image : public rclcpp::Node
...
+  }
+
+  IMAGE_TOOLS_LOCAL
+  bool help(const std::vector<std::string> args)
+  {
+    if (std::find(args.begin(), args.end(), "--help") != args.end() ||
+      std::find(args.begin(), args.end(), "-h") != args.end())
+    {
+      std::stringstream ss;
+      ss << "Usage: cam2image [-h] [--ros-args [-p param:=value] ...]" << std::endl;
```
dirk-thomas: Nitpick: each `std::endl` also flushes the stream which doesn't seem necessary on each line.

jacobperron:
> I was hoping the nodes would declare their parameters and --help would then call a generic function which get those descriptors and renders the usage.

It seemed easier to do something like was done before to ensure formatting looked okay.
But, this is a nice idea. I think that a generic function like, `get_node_usage()` should live in the client library or a utility somewhere that node developers can use. I can open up a feature request for this.

PR #402, opened 2019-10-18
Refactor intra-process demo into components.

hidmic:
Precisely what the title says.


hidmic:
@mjcarroll I migrated the `two_node_pipeline` demo, but not the rest. I was unsure to what extent we wanted to migrate. WDYT?

PR #408, opened 2019-10-24
Don't redefine add_dependencies

rotu:
rosidl defines `add_dependencies`, and cmake will gladly redefine functions without a warning. If you have a multi-project CMake tree, this can lead to weirdness

CMake Error at src/ros2/demos/demo_nodes_cpp/CMakeLists.txt:42 (target_compile_definitions):
  target_compile_definitions called with non-compilable target type
Call Stack (most recent call first):
  install/share/rosidl_typesupport_cpp/cmake/rosidl_typesupport_cpp_generate_interfaces.cmake:129 (add_dependencies)
  install/share/ament_cmake_core/cmake/core/ament_execute_extensions.cmake:48 (include)
  install/share/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake:286 (ament_execute_extensions)
  src/ros2/demos/logging_demo/CMakeLists.txt:21 (rosidl_generate_interfaces)

Signed-off-by: Dan Rose <dan@digilabs.io>

### DIFF
--- a/demo_nodes_cpp/CMakeLists.txt
+++ b/demo_nodes_cpp/CMakeLists.txt
@@ -32,7 +32,7 @@ function(custom_executable subfolder target)
   DESTINATION lib/${PROJECT_NAME})
 endfunction()
 
-function(add_dependencies library)
+function(add_demo_dependencies library)
   target_compile_definitions(${library}
     PRIVATE "DEMO_NODES_CPP_BUILDING_DLL")
   ament_target_dependencies(${library}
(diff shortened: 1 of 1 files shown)


rotu:
Btw, @dirk-thomas, this might be interesting to you. I didn't realize until this point that functions and macros are defined globally. So while you can collect multiple projects with `add_subdirectory` (e.g. https://gist.github.com/rotu/1eac858b808b82bbf1b475f515e91636), I think it's too much to ask to expect the *entire ROS2 tree* to build in this way.

dirk-thomas:
> I think it's too much to ask to expect the entire ROS2 tree to build in this way.

That is the reason we have dropped this compared to ROS 1 / catkin (see http://design.ros2.org/articles/ament.html#building-within-a-single-cmake-context).

dirk-thomas:
Thanks for the patch.

rotu:
Thanks for the background! This has turned out to be a very educational misadventure

Issue #292, opened 2018-12-11
[allocator_tutorial] More global allocations used when intra-process is enabled using Fast_RTPS


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Bionic
- Installation type:
  - debs
- Version or commit hash:
  - crystal pre-release
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - 

#### Steps to reproduce issue

```

```

#### Additional information

----

sloretz:
Labeled as question because I'm not sure what the correct expected behavior is, but the actual behavior is unexpected to me.

wjwwood:
I don't know for sure what should be expected here, but I can say that the intra process implementation does do memory allocations depending on what signature of publish is called by the user.

PR #409, opened 2019-10-24
Add in a more helpful usage message to allocator_tutorial.

clalancette:

### DIFF
--- a/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
+++ b/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
@@ -143,6 +143,14 @@ int main(int argc, char ** argv)
   std::list<std::string> keys = {"intra", "intraprocess", "intra-process", "intra_process"};
   bool intra_process = false;
 
+  printf("This simple demo shows off a custom memory allocator to count all\n"
+    "instances of new/delete in the program.  It can be run in either regular\n"
+    "mode (no arguments), or in intra-process mode (by passing 'intra' as a\n"
+    "command-line argument)'.  It will then publish a message to the\n"
+    "'/allocator_tutorial' topic every 10 milliseconds until Ctrl-C is pressed.\n"
+    "At that time it will print a count of the number of allocations and\n"
+    "deallocations that happened during the program.\n\n");
+
   if (argc > 1) {
     for (auto & key : keys) {
       if (std::string(argv[1]) == key) {


demo_nodes_cpp/src/topics/allocator_tutorial.cpp
```diff
@@ -143,6 +143,14 @@ int main(int argc, char ** argv)
   std::list<std::string> keys = {"intra", "intraprocess", "intra-process", "intra_process"};
   bool intra_process = false;
 
+  printf("This simple demo shows off a custom memory allocator to count all\n"
+    "instances of new/delete in the program.  It can be run in either regular\n"
+    "mode (no arguments), or in intra-process mode (by passing 'intra' as a\n"
+    "command-line argument)'.  It will then publish a message to the\n"
+    "'/allocator_tutorial' topic every 10 milliseconds until Ctrl-C is pressed.\n"
+    "At that time it will print a count of the number of allocations and\n"
+    "deallocations that happened during the program.\n\n");
```
hidmic: @clalancette why not turning this into a usage message, only shown if `--help` or `-h` are present?

clalancette: The main reason I wanted to print this at startup time is that running the binary is quite silent:

```
ros2 run demo_nodes_cpp allocator_tutorial
```

(it only prints a single debugging message).  Unless you go look at the code, it isn't clear at all what this tutorial is supposed to do, or what users should expect to happen.  So I thought that unconditionally printing this message would help future testers not ask the same question.

mjcarroll: Could we maybe say what the expected results would be?  Like if you are using `intra` you would expect allocations to be lower? 

Since this is just a demo, I feel like erring on the side of more verbosity is quite alright.

clalancette: I just tested out the difference between running with 'intra' and without.  I ran the tutorial for 10 seconds, using the following command:

```
timelimit -t 10 -s 2 install/demo_nodes_cpp/lib/demo_nodes_cpp/allocator_tutorial
```

This causes the tutorial to generate on average 4089 calls to global new and 13522 calls to allocator new.

If I instead run:

```
timelimit -t 10 -s 2 install/demo_nodes_cpp/lib/demo_nodes_cpp/allocator_tutorial intra
```

This cause the tutorial to generate on average 3910 calls to global new and 13588 calls to allocator new.  So with intra-process there are fewer calls to the global one, but there are *more* calls to the allocator one.  Also, the difference is probably in the noise, since I wasn't on a "quiet" system.

Thus, I don't think we can say too much about the allocations.  I'd say we can probably go forward with this as-is.

I'll run CI on it just to make sure I didn't mess something obvious up.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8644/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4487/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7051/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8563/)

demo_nodes_cpp/src/topics/allocator_tutorial.cpp
sloretz: > Thus, I don't think we can say too much about the allocations.

Related ticket: ros2/demos#292

clalancette:
@mjcarroll This is ready to go; I leave it to you whether you want to merge this for Eloquent or wait.

PR #410, opened 2019-10-24
Remove the warning about Fast-RTPS not supporting QoS Liveliness.

clalancette:
As far as I can tell, it now does.

@emersonknapp The tests seem to work for me on Fast-RTPS (albeit with some warnings).  Can you confirm if it is correct to state that Fast-RTPS now supports liveliness?


emersonknapp:
Yes, it is correct to state that :) we have full liveliness support from FastRTPS

clalancette:
> Yes, it is correct to state that :) we have full liveliness support from FastRTPS

Great, thanks!

clalancette:
I'm going to merge this despite the failing CI build, since it is clearly not caused by this.

PR #413, opened 2019-11-05
Remove unnecessary dependency on ros2run

hidmic:
Precisely what the title says.

### DIFF
--- a/dummy_robot/dummy_robot_bringup/package.xml
+++ b/dummy_robot/dummy_robot_bringup/package.xml
@@ -17,7 +17,6 @@
   <exec_depend>launch</exec_depend>
   <exec_depend>launch_ros</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
-  <exec_depend>ros2run</exec_depend>
 
   <test_depend>ament_cmake_gtest</test_depend>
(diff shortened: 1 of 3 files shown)


dirk-thomas:
In order to remove the dependency you must also remove its usage from here: https://github.com/ros2/demos/blob/bde11a9f7d2cd4504feb430920caec52a47dff56/pendulum_control/test/execute_with_delay.py#L19

hidmic:
Argh, I had not seen that use. Brought back in https://github.com/ros2/demos/pull/413/commits/1665dbc170f4e58c8cd263e21e288608e4f33a53.

hidmic:
Sanity CI for `lifecycle` and `dummy_robot_bringup`:
  * Linux [Build Status](http://ci.ros2.org/job/ci_linux/8607/)

dirk-thomas:
I though the goal was to remove all dependencies on the `ros2cli` repo?

hidmic:
> I though the goal was to remove all dependencies on the `ros2cli` repo?

We would have to get rid of the `ros_testing` -> `ros2test` -> `ros2cli` chain of dependencies, so it doesn't seem possible (unless we re-discuss that as well). Anyways, unnecessary dependencies can still go away.

PR #423, opened 2019-12-18
Add try/catch block to fix process hang

### LINKED ISSUE #407: QoS_demo_cpp: Process hung on [INFO] [rclcpp]: signal_handler(signal_value=2)
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04 
- Installation type:
  - Source 
- Version or commit hash:
  - Eloquent 
- DDS implementation:
  - Connext 
  - rclcpp 

#### Steps to reproduce issue

```
RMW_IMPLEMENTATION=rmw_connext_cpp ros2 run quality_of_service_demo_cpp interactive_publisher --delay 0.5 --deadline 1 --liveliness MANUAL_BY_TOPIC --lease 1
RMW_IMPLEMENTATION=rmw_connext_cpp ros2 run quality_of_service_demo_cpp interactive_subscriber --deadline 1 --liveliness MANUAL_BY_TOPIC --lease 1
```

#### Expected behavior
print
```
^C[INFO] [rclcpp]: signal_handler(signal_value=2)
```
then exit the process
#### Actual behavior
```
[INFO] [rclcpp]: signal_handler(signal_value=2)
[INFO] [rclcpp]: signal_handler(signal_value=2)
[INFO] [rclcpp]: signal_handler(signal_value=2)
[INFO] [rclcpp]: signal_handler(signal_value=2)
```
never exit.

claireyywang:
Fix #407

### DIFF
--- a/quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
+++ b/quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
@@ -157,15 +157,20 @@ int main(int argc, char * argv[])
 
   PublisherCommandHandler cmd_handler(exec, talker);
 
-  talker->initialize();
-  talker->print_qos();
+  try
+  {
+    talker->initialize();
+    talker->print_qos();
 
-  cmd_handler.start();
-  exec.add_node(talker);
-  exec.spin();
-  exec.remove_node(talker);
-  cmd_handler.stop();
-
+    cmd_handler.start();
+    exec.add_node(talker);
+    exec.spin();
+  }
+  catch (int n)
+  {
+    exec.remove_node(talker);
+    cmd_handler.stop();
+  }
   rclcpp::shutdown();
 
   return 0;
--- a/quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
+++ b/quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
@@ -166,7 +166,7 @@ int main(int argc, char * argv[])
     exec.add_node(talker);
     exec.spin();
   }
-  catch (int n)
+  catch (const std::exception&)
   {
     exec.remove_node(talker);
     cmd_handler.stop();
--- a/quality_of_service_demo/rclcpp/src/interactive_subscriber.cpp
+++ b/quality_of_service_demo/rclcpp/src/interactive_subscriber.cpp
@@ -134,16 +134,20 @@ int main(int argc, char * argv[])
     };
 
   SubscriberCommandHandler cmd_handler(exec, listener);
+  try
+  {
+    listener->initialize();
+    listener->print_qos();
 
-  listener->initialize();
-  listener->print_qos();
-
-  cmd_handler.start();
-  exec.add_node(listener);
-  exec.spin();
-  exec.remove_node(listener);
-  cmd_handler.stop();
-
+    cmd_handler.start();
+    exec.add_node(listener);
+    exec.spin();
+  }
+  catch (int n)
+  {
+    exec.remove_node(listener);
+    cmd_handler.stop();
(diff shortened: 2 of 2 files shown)


quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
```diff
@@ -157,15 +157,20 @@ int main(int argc, char * argv[])
...
+  try
+  {
+    talker->initialize();
+    talker->print_qos();
 
+    cmd_handler.start();
+    exec.add_node(talker);
+    exec.spin();
+  }
+  catch (int n)
```
ivanpauno: Sorry, I don't understand the fix.
Where this exception was thrown?

claireyywang: I'm trying to use a generic exception, but I guess `int n` is not the right one. I have updated it to `const std::exception&`.

ivanpauno: You can just use:
```cpp
} catch (...) {
```

Though catching particularly `std::exception` allows you to print a better error message (exceptions following the `std::exception` interface have a [what](https://en.cppreference.com/w/cpp/error/exception/what) method).
Sometimes, when people want to catch everything and provide as much information as possible they do something like [this](https://github.com/ros2/rclcpp/blob/0c66d0c725ad1f977e2b907bc0619d6ceb6de238/rclcpp/src/rclcpp/context.cpp#L47-L51).

----

I still don't see how this solves #407, can you explain further?

claireyywang: I'm trying to mimic the behaviour of `try/except` block like in this python [example](https://github.com/ros2/demos/blob/master/demo_nodes_py/demo_nodes_py/topics/listener.py#L35). I think by treating `ctrl-c` as a keyboard exception, the process is able to run the `exec.remove_node` and `cmd_handler.stop` immediately instead of waiting for the separate thread. After testing it on ubuntu 18.04, it seems to work. Thus I made the PR. There could be a better way to solve this problem. This is just my attempt to pitch something that worked for me.

claireyywang:
Linux 
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4705/)
OSX 
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8841/)

quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
ivanpauno: cpp doesn't throw an exception when treating ctrl-c, does something Python specific.
If you print something in the catch statement, you probably won't see anything.

The only way of catching a signal in cpp is by installing a signal handler, and as `rclcpp` already installs one, you will have to do chaining. See https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/signal_handler.cpp.

hidmic: @ivanpauno is correct. This patch as-is won't do.

claireyywang: I have backlogged this PR cuz I don't have time to work on it in recent weeks, and the process can be killed if use `ctrl-c` and `x` as discussed in the original issue. I'll come back to it at some point...

hidmic:
I suspect no one's picking this up any time soon @mabelzhang @mjeronimo. Shall we close?

mabelzhang:
No objections to closing.

mjeronimo:
No objections here either. Closing.

PR #425, opened 2020-01-14
Added more parameters for camera topic examples.

DanBmh:
Added parameter for custom topic names, display window name and camera device id.

Signed-off-by: DanBmh


DanBmh:
I didnt think about remapping before, but will this still work if i have two cam2img nodes? In my setup i have two cameras and i would like to stream both of them to their own topic.

Regarding the window name i think i will get the same problem of two windows with the same name if i want to display both camera streams.

Also i dont think the extra parameters will distract users much from the demo as they are optional, but they make the demo (as a build in feature) more usable for other purposes. (I could replace an old ros1 package i used before)

By the way, i was very glad when i found out yesterday that the topic argument existed before.  i have ros2 crystal running on a raspi and could not build the package with my changes as a lot other things here have changed inbetween.

jacobperron:
> I didnt think about remapping before, but will this still work if i have two cam2img nodes? In my setup i have two cameras and i would like to stream both of them to their own topic.

Yes, remapping can be configured per node. For example, if there is more than one node in a process, then we can prefix remaps with the node name and a colon ([reference](https://design.ros2.org/articles/static_remapping.html#supporting-remap-one-node-in-a-process)):

    ros2 run image_tools showimage --ros-args -r showimage:image:=another_topic

Alternatively, launch supports remapping node topics (e.g. see [launch XML design](https://design.ros2.org/articles/roslaunch_xml.html#remap-tag))

I'll add that this demo code is just that, a demonstration. It is not intended to be used directly in other applications and so I'm not convinced we need to worry about supporting more general use-cases. I recommend adapting the code as you need for your application, or use existing tools like [`rqt_image_view`](https://github.com/ros-visualization/rqt_image_view) for viewing images from ROS topics.

DanBmh:
Thanks for the example and links. If i try this with one node it is working. But if i try it with two nodes it doesnt do most of the times. Using following commands:

```
sleep 1; ros2 run image_tools cam2image --ros-args -r __ns:=/card_cam_node -p device_id:=0
sleep 2; ros2 run image_tools cam2image --ros-args -r __ns:=/face_cam_node -p device_id:=1
```

i think the same problem would occur with my topic arguments, as i believe the problem is that im using the same node twice. I also tried to rename the node only with `-r __node:=card_cam_node` and to remap the topic `-r cam2img:image:=card_cam` but this doesnt work too. I always get only the topics of the node i started last.

But in rare cases i got both nodes topics at the same time, as intended, but using the same commands and order again (im starting it with docker-compose), the next time i only get one nodes topics. 

Is there an argument to run both nodes at the same time?
And do you know the reason for the behavior that my commands did work sometimes?

jacobperron:
`ros2 run` is a blocking call. Can you try running the commands in separate shells? I suggest you try using the launch system for starting multiple nodes at the same time. For example, the following launch file should work in Eloquent:

```xml
<launch>
  <node node-name="card_cam" pkg="image_tools" exec="cam2image" output="screen">
    <param name="device_id" value="0" />
  </node>
  <node node-name="face_cam" pkg="image_tools" exec="cam2image" output="screen">
    <param name="device_id" value="1" />
  </node>
</launch>
```

You can start it with `ros2 launch`.

image_tools/src/cam2image.cpp
```diff
@@ -181,6 +180,8 @@ class Cam2Image : public rclcpp::Node
       ss << std::endl;
       ss << "  show_camera\tShow camera stream. Either 'true' or 'false' (default)";
       ss << std::endl;
+      ss << "  device_id\tDevice id of the camera. Default is 0 for camera video0";
```
jacobperron:
```suggestion
      ss << "  device_id\tDevice ID of the camera. 0 (default) selects the default camera device."';
```

image_tools/src/cam2image.cpp
```diff
@@ -300,6 +302,7 @@ class Cam2Image : public rclcpp::Node
   size_t width_;
   size_t height_;
   bool burger_mode_;
+  size_t device_id_;
```
jacobperron: Since opencv accepts an `int` ([reference](https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html#a5d5f5dacb77bbebdcbfb341e3d4355c1)).
```suggestion
  int device_id_;
```

image_tools/src/showimage.cpp
```diff
@@ -206,8 +209,8 @@ class ShowImage : public rclcpp::Node
 
       cv::Mat cvframe = frame;
 
-      // Show the image in a window called "showimage".
-      cv::imshow("showimage", cvframe);
+      // Show the image in a window
+      cv::imshow(window_name_, cvframe);
```
jacobperron: IMO, it would be more elegant to use the topic name as the window name. This has the effect the number of windows will equal the number of image topics being subscribed to by all `showimage` nodes. It also avoids a user forgetting to set the window name parameter.

I think this should work (I haven't tested it):

```suggestion
      cv::imshow(sub_->get_topic_name(), cvframe);
```

DanBmh:
I added your suggestion to use the topic name as window name, but only as default value so that a user still can use a custom window name if he likes to.

Using a launch file to start the two camera nodes did work for me. Afterwards i tried running the two separate commands again and i found out this only works if i dont give parameters with -p. If i run it with extra parameters like frequency or width and height the topics of one node always were missing. So im using the launch file solution now.

image_tools/src/showimage.cpp
```diff
@@ -82,6 +77,11 @@ class ShowImage : public rclcpp::Node
 
     RCLCPP_INFO(this->get_logger(), "Subscribing to topic '%s'", topic_.c_str());
     sub_ = create_subscription<sensor_msgs::msg::Image>(topic_, qos, callback);
+
+    if(window_name_ == ""){
```
jacobperron: There's an uncrustify error here:
```suggestion
    if (window_name_ == "") {
```

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9240/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4927/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7548/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9159/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/40/)

jacobperron:
There was a compiler warning on Windows, I think 3ba0218 should fix it.
Windows: [Build Status](https://ci.ros2.org/job/ci_windows/9160/)

DanBmh:
What do we do with this warning now?

jacobperron:
> What do we do with this warning now?

Sorry, this PR dropped off my radar. I can spend some time investigating on a Windows host when I have a moment.

jacobperron:
@DanBmh Sorry, I accidentally pushed the wrong changes to your fork when trying to rebase this PR. I've reopened a PR with your changes in https://github.com/ros2/demos/pull/465. I think I've resolved the compiler warning on Windows too.

PR #427, opened 2020-01-29
Use `spin_until_future_complete` instead of `spin_some` in parameters_event demo

ivanpauno:
From https://github.com/ros2/rclcpp/pull/844, `spin_some` collects pending executables just once, and executes all of them.

That broke this example, because the following happens.
The clients requests aren't handled by the services until `spin` is called.
The parameter event is published by the service when handling the requests.

If executables are not collected more than once, `on_parameter_event` callbacks won't be ever called (the only ones called are the ones from the previous `declare_parameter` calls).

The bad thing, is that now the user have to terminate the example by using `ctrl-c` (maybe, add a log saying that).

Fixes: https://ci.ros2.org/view/nightly/job/nightly_linux_release/1428/testReport/junit/demo_nodes_cpp/TestExecutablesTutorial/test_processes_output/.

### DIFF
--- a/demo_nodes_cpp/src/parameters/parameter_events.cpp
+++ b/demo_nodes_cpp/src/parameters/parameter_events.cpp
@@ -88,12 +88,7 @@ int main(int argc, char ** argv)
 
   // TODO(wjwwood): Create and use delete_parameter
 
-  // TODO(hidmic): Fast-RTPS takes a significant amount of time to deliver
-  //               requests and response, thus the rather long sleep. Reduce
-  //               once that's resolved.
-  rclcpp::sleep_for(3s);
-
-  rclcpp::spin_some(node);
+  rclcpp::spin(node);
 
   rclcpp::shutdown();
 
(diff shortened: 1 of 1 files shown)


wjwwood:
Or you could spin in a thread, and call shutdown after some condition is met, if you want to avoid a future.

ivanpauno:
> I don't think we should change the behavior of this demo (now it will block until ctrl-c where as before it would close on its own).
>
> Instead, we should setup a future, complete it based on some condition, and use > spin_until_future_complete.

I also thought about that, but it makes the example more complex.
The only condition I can think is: the callback was called 10 times.

Changes requires: https://github.com/ros2/rclcpp/pull/971

ivanpauno:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/9178/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4884/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/7486/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9089/)

PR #428, opened 2020-01-29
Add gitignore file

ivanpauno:
I've had to skip `__pycache__` folders many time in this repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Byte compiled Python
+__pycache__/


dirk-thomas:
If you are developing Python code I would highly suggest to add this to your local `.gitignore` file.

ivanpauno:
We do have in other repos a `.gitignore` ignoring this, e.g. https://github.com/ros2/rclpy/blob/30eeae9b30ab15209798d62b501d1dec633e1db3/.gitignore#L2.
I think that adding a `.gitignore` including what has to be ignored in the normal workflow when using the repo is a good idea (`colcon build --symlink-install` generates those folders).

> If you are developing Python code I would highly suggest to add this to your local .gitignore file.

Yes, I can do that too.

----

If we don't want to include a `.gitignore` in all the repos, we can close this PR.

PR #436, opened 2020-04-14
Primary state error transitions

thebyohazard:
This PR shows an example of the new self-transitioning functionality implemented in required PRs [ros2/rcl_interfaces#97](https://github.com/ros2/rcl_interfaces/pull/97), [ros2/rcl#618](https://github.com/ros2/rcl/pull/618), and [ros2/rclcpp#1064](https://github.com/ros2/rclcpp/pull/1064). Ideally, I think the lifecycle_service_client should react to the lifecycle_talker having an error rather than just knowing it will happen as that would be more true to a real world example, but for a demo and proof of concept, maybe it's better to keep it simpler.


lifecycle/src/lifecycle_talker.cpp
```diff
@@ -88,6 +88,15 @@ class LifecycleTalker : public rclcpp_lifecycle::LifecycleNode
         get_logger(), "Lifecycle publisher is active. Publishing: [%s]", msg->data.c_str());
     }
 
+    // Simulate the node having some error whilst active,
+    // necessitating reconfiguration.
+    // raise_error() can be called from an inactive or active state.
+    if(count == 32){
+        RCLCPP_ERROR(get_logger(),"An error has arisen in the node and it needs reconfiguration.");
+        raise_error();
```
fujitatomoya, line 96: check if the current state in ACTIVE before faise_error() ?

lifecycle/src/lifecycle_talker.cpp
```diff
@@ -262,6 +271,40 @@ class LifecycleTalker : public rclcpp_lifecycle::LifecycleNode
...
+   * enters the "errorProcessing" state.
+   * Depending on the return value of this function, the state machine
+   * either invokes a transition to the "unconfigured" state or the
+   * finalized state.
+   * TRANSITION_CALLBACK_SUCCESS transitions to "unconfigured"
+   * TRANSITION_CALLBACK_FAILURE transitions to "finalized"
+   * TRANSITION_CALLBACK_ERROR or any uncaught exceptions to "finalized"
+   */
+    rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+  on_error(const rclcpp_lifecycle::State & state)
```
fujitatomoya, line 286: indent is not correct?

thebyohazard: Thanks for that. Fixed

norro:
What is the state of this PR?

fujitatomoya:
not sure, @thebyohazard any update you would do?

PR #440, opened 2020-04-19
Add XML and YAML launch scripts for dummy_robot_bringup

### LINKED ISSUE #434: Use launch front-end for dummy_robot_bringup
#### Feature description

As of Eloquent, we can use XML and YAML markup for launch scripts. I think it would be nice to launch scripts using one (or both) of the markup languages, that can replace or sit alongside the [existing launch file](https://github.com/ros2/demos/blob/master/dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup.launch.py).

p-vega:
Add XML and YAML launch scripts for dummy_robot_bringup
Fixes #434 
Signed-off-by: Pablo Vega <epvega@gmail.com>

### DIFF
--- a/dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup_launch.xml
+++ b/dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup_launch.xml
@@ -1,0 +1,8 @@
+<launch>
+  <node pkg="dummy_map_server" exec="dummy_map_server" output="screen" />
+  <node pkg="robot_state_publisher" exec="robot_state_publisher" output="screen" >
+    <param name='robot_description' value="$(command 'cat $(find-pkg-share dummy_robot_bringup)/launch/single_rrbot.urdf')"/>
+  </node>
+  <node pkg="dummy_sensors" exec="dummy_joint_states" output="screen"/>
+  <node pkg="dummy_sensors" exec="dummy_laser" output="screen"/>
+</launch>
--- a/dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup_launch.xml
+++ b/dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup_launch.xml
@@ -1,7 +1,7 @@
 <launch>
   <node pkg="dummy_map_server" exec="dummy_map_server" output="screen" />
   <node pkg="robot_state_publisher" exec="robot_state_publisher" output="screen" >
-    <param name='robot_description' value="$(command 'cat $(find-pkg-share dummy_robot_bringup)/launch/single_rrbot.urdf')"/>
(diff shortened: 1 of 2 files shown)


dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup_launch.yaml
```diff
@@ -0,0 +1,18 @@
+launch:
+- node:
+    pkg: dummy_map_server
+    exec: dummy_map_server
+    output: screen
+- node:
+    pkg: robot_state_publisher
+    exec: robot_state_publisher
+    output: screen
+    args: $(find-pkg-share dummy_robot_bringup)/launch/single_rrbot.urdf
```
hidmic, line 10: @p-vega this form of `robot_state_publisher` configuration [is deprecated](https://github.com/ros/robot_state_publisher/blob/5b3900eb554589b6bf1499cbca3ec35b9899567f/src/robot_state_publisher.cpp#L89-L93) -- use parameters instead of command line arguments, like in the XML.

p-vega: @hidmic Thanks for reviewing, I couldn't find any example of the proper parameter syntax for the yaml launch script. Can you point me in the right direction?

hidmic: Sure, check [this test](https://github.com/ros2/launch_ros/blob/master/test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py).

dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup_launch.xml
```diff
@@ -0,0 +1,8 @@
+<launch>
+  <node pkg="dummy_map_server" exec="dummy_map_server" output="screen" />
+  <node pkg="robot_state_publisher" exec="robot_state_publisher" output="screen" >
+    <param name='robot_description' value="$(command 'cat $(find-pkg-share dummy_robot_bringup)/launch/single_rrbot.urdf')"/>
```
hidmic, line 4: @p-vega why single quotes for the `name` attribute value?

jacobperron:
> shall we merge this now or defer it till after Foxy?

@hidmic I don't think we need to defer since this is a leaf package.

dummy_robot/dummy_robot_bringup/launch/dummy_robot_bringup_launch.xml
p-vega: @hidmic you're right, they should be double quotes. I'm sorry for the typo, I'll change it and I'll add a new commit.

hidmic:
Alright, there're no tests covering this functionality so I don't think running CI's worth it. Merging.

Thank you for your contribution @p-vega !

PR #456, opened 2020-06-23
Add missing required parameter in LifecycleNode launch action

ivanpauno:
See https://github.com/ros2/launch_ros/pull/157.

### DIFF
--- a/lifecycle/launch/lifecycle_demo.launch.py
+++ b/lifecycle/launch/lifecycle_demo.launch.py
@@ -20,7 +20,7 @@ from launch_ros.actions import Node
 def generate_launch_description():
     return LaunchDescription([
         LifecycleNode(package='lifecycle', executable='lifecycle_talker',
-                      name='lc_talker', output='screen'),
+                      name='lc_talker', namespace='', output='screen'),
         Node(package='lifecycle', executable='lifecycle_listener', output='screen'),
         Node(package='lifecycle', executable='lifecycle_service_client', output='screen')
     ])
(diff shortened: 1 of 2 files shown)


lifecycle/launch/lifecycle_demo.launch.py
```diff
@@ -20,7 +20,7 @@
 def generate_launch_description():
     return LaunchDescription([
         LifecycleNode(package='lifecycle', executable='lifecycle_talker',
-                      name='lc_talker', output='screen'),
+                      name='lc_talker', namespace='', output='screen'),
```
sloretz, line 23: Is `namespace=''` the same as `namespace='/'`? If so, the latter seems clearer to me

ivanpauno: It's the same in this case.
The difference is that `''` can be modified when the launch file is being included in another launch file.

```python3
LaunchDescription([
  ...
  PushRosNamespace('my_ns'),
  IncludeLaunchDescription(path_to_lifecycle_demo_launch_py),
  ...
])
```

See examples of the difference here https://github.com/ros2/launch_ros/blob/master/test_launch_ros/test/test_launch_ros/actions/test_push_ros_namespace.py.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11261/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6521/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9220/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11169/)

ivanpauno:

PR #467, opened 2020-10-26
Implement get_rcl_allocator.

stevewolter:
This is needed as a companion change to rclcpp #1324.

Signed-off-by: Steve Wolter <swolter@google.com>

### DIFF
--- a/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
+++ b/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
@@ -92,6 +92,33 @@ constexpr bool operator!=(
   return false;
 }
 
+// You need to overload get_rcl_allocator for custom allocators.
+// This function needs to build an instance of rcl_allocator_t
+// for use in C code.
+template<typename T>
+rcl_allocator_t get_rcl_allocator(MyAllocator<T>)
+{
+  rcl_allocator_t rcl_allocator;
+  rcl_allocator.allocate = [](size_t size, void *) {
+      num_allocs++;
+      return malloc(size);
+    };
+  rcl_allocator.deallocate = [](void * pointer, void *) {
+      num_deallocs++;
+      return free(pointer);
+    };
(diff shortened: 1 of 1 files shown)


demo_nodes_cpp/src/topics/allocator_tutorial.cpp
```diff
@@ -92,6 +92,33 @@ constexpr bool operator!=(
   return false;
 }
 
+// You need to overload get_rcl_allocator for custom allocators.
+// This function needs to build an instance of rcl_allocator_t
+// for use in C code.
+template<typename T>
+rcl_allocator_t get_rcl_allocator(MyAllocator<T>)
```
wjwwood, line 99: Doesn't this need to be in the `rclcpp` namespace?

stevewolter: Argument-dependent lookup saves us: https://en.cppreference.com/w/cpp/language/adl. As long as the function is in the namespace of one of its arguments, the overload will be resolved.

wjwwood: Yes, but should it be in the same namespace... I feel like the answer is yes, but I'm open to suggestion. I'm thinking from the perspective of a person reading this code and not knowing what it is exactly that is being overridden or where it comes from. The namespace gives important context imo.

stevewolter: Got you. I don't care one way or another, so if you stumble over the namespace, it's probably a good signal that many devs would stumble over the namespace. Moved it to rclcpp.

rclcpp::allocator wouldn't work because the calling code is in rclcpp.

wjwwood: It should be in the `rclcpp::allocator` namespace, I commented on the original pr: https://github.com/ros2/rclcpp/pull/1324#pullrequestreview-517938632

PR #468, opened 2020-10-26
Update deprecated qos policy value names

ivanpauno:
See https://github.com/ros2/rclpy/pull/634.

### DIFF
--- a/demo_nodes_cpp/src/topics/listener.cpp
+++ b/demo_nodes_cpp/src/topics/listener.cpp
@@ -43,12 +43,10 @@ public:
     // Note that not all publishers on the same topic with the same type will be compatible:
     // they must have compatible Quality of Service policies.
     sub_ = create_subscription<std_msgs::msg::String>("chatter", 10, callback);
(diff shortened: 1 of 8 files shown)


demo_nodes_cpp/src/topics/listener.cpp
```diff
@@ -43,10 +43,12 @@ class Listener : public rclcpp::Node
     // Note that not all publishers on the same topic with the same type will be compatible:
     // they must have compatible Quality of Service policies.
     sub_ = create_subscription<std_msgs::msg::String>("chatter", 10, callback);
+    sub2_ = create_subscription<std_msgs::msg::String>("chatter", rclcpp::QoS(5).best_effort(), callback);
```
jacobperron, line 46: This change looks unrelated to updating the deprecated usage.

ivanpauno: ups, that was an unrelated local change that I didn't realize I was pushing.

ivanpauno: Fixed in https://github.com/ros2/demos/pull/468/commits/3c478946a85df00a818d5c02127db67b0bc934e2.

jacobperron (approved):

PR #492, opened 2021-03-29
Fix integer type in RCLCPP_* macro printf.

clalancette:
While building on macOS, a warning shows up like:

format specifies type 'long' but the argument has type 'int64_t' (aka 'long long')

The problem is that %ld isn't correct for int64_t on all
platforms.  Switch to the PRId64 macro so the type is correct
everywhere.

### DIFF
--- a/demo_nodes_cpp/src/parameters/parameter_event_handler.cpp
+++ b/demo_nodes_cpp/src/parameters/parameter_event_handler.cpp
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cinttypes>
 #include <memory>
 #include <regex>
 #include <string>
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14112/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8920/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11801/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14208/)

mjeronimo (approved):

clalancette:
The failure on Windows is not because of this PR.  The failure on PR jobs seems to be a long-standing issue.  I'm going to go ahead and merge this, thanks for the review.

PR #494, opened 2021-04-01
change ParameterEventHandler to take events as const ref instead of shared pointer

wjwwood:
This is related to, and requires, https://github.com/ros2/rclcpp/pull/1598

@mjeronimo can you think of any other places we need to update something like this? Documentation?

### DIFF
--- a/lifecycle/CMakeLists.txt
+++ b/lifecycle/CMakeLists.txt
@@ -17,34 +17,27 @@ find_package(lifecycle_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(std_msgs REQUIRED)
 
-include_directories(
-  include
-  ${std_msgs_INCLUDE_DIRS}
-  ${lifecycle_msgs_INCLUDE_DIRS}
-  ${rclcpp_lifecycle_INCLUDE_DIRS}
-  ${rclcpp_INCLUDE_DIRS})
-
 ### demos
 add_executable(lifecycle_talker
   src/lifecycle_talker.cpp)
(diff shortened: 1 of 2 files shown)


lifecycle/CMakeLists.txt
```diff
@@ -30,18 +30,21 @@ add_executable(lifecycle_talker
...
   ${rclcpp_lifecycle_LIBRARIES}
   ${std_msgs_LIBRARIES}
+  rclcpp::rclcpp
 )
 add_executable(lifecycle_service_client
   src/lifecycle_service_client.cpp)
 target_link_libraries(lifecycle_service_client
   ${rclcpp_lifecycle_LIBRARIES}
   ${std_msgs_LIBRARIES}
+  rclcpp::rclcpp
```
wjwwood, line 47: Instead, let's try using `ament_target_dependencies`, like our current template suggests:

https://github.com/ros2/ros2cli/blob/d03ee6ce090b845ffac64af9ca070c1f71460743/ros2pkg/ros2pkg/resource/ament_cmake/CMakeLists.txt.em#L65-L70

mjeronimo:
Yes, probably need to update the ROS2 tutorial for the ParameterEventHandler. I'll take a look.

mjeronimo (approved):

mjeronimo:
As it turns out the ParameterEventHandler tutorial doesn't need to be updated as it doesn't use the add_parameter_event_handler() method (which is the one use to monitor all parameter events); instead, it uses add_parameter_handler(), which is unaffected by the change from the client code point of view.

PR #495, opened 2021-04-02
Depend on rclcpp::rclcpp target

audrow:
This PR extends #494.


audrow:
Actually, I realized this will be overwritten when I run it in CI with my current setup. I'll push these changes to #494 instead.

PR #497, opened 2021-04-06
change how serialized message works with subscription

wjwwood:
Connected to https://github.com/ros2/rclcpp/pull/1557

### DIFF
--- a/demo_nodes_cpp/src/topics/listener_serialized_message.cpp
+++ b/demo_nodes_cpp/src/topics/listener_serialized_message.cpp
@@ -70,7 +70,7 @@ public:
   }
 
 private:
-  rclcpp::Subscription<rclcpp::SerializedMessage>::SharedPtr sub_;
+  rclcpp::Subscription<std_msgs::msg::String>::SharedPtr sub_;
 };
 
 }  // namespace demo_nodes_cpp


audrow (approved):

audrow:
Here's CI (green): https://github.com/ros2/rclcpp/pull/1557#issuecomment-833990334

demo_nodes_cpp/src/topics/listener_serialized_message.cpp
```diff
@@ -70,7 +70,7 @@ class SerializedMessageListener : public rclcpp::Node
   }
 
 private:
-  rclcpp::Subscription<rclcpp::SerializedMessage>::SharedPtr sub_;
+  rclcpp::Subscription<std_msgs::msg::String>::SharedPtr sub_;
```
ivanpauno, line 73: I think it's worth to add a note in the release notes about this API change: https://github.com/ros2/ros2_documentation/blob/rolling/source/Releases/Release-Humble-Hawksbill.rst

wjwwood: I agree, at the time we were waiting on that page to be created. @audrow does it make sense to add it there now?

PR #500, opened 2021-04-16
Small fixes for even_parameters_node.

clalancette:
First, make it print something on startup so that the user
has some idea of what it is supposed to do.  Second, make
sure to set the reason for failure if we fail to set a
parameter.

### DIFF
--- a/demo_nodes_cpp/src/parameters/even_parameters_node.cpp
+++ b/demo_nodes_cpp/src/parameters/even_parameters_node.cpp
@@ -28,12 +28,10 @@ public:
   explicit EvenParameterNode(rclcpp::NodeOptions options)
   : Node("even_parameters_node", options.allow_undeclared_parameters(true))
   {
-    // Force flush of the stdout buffer.
-    setvbuf(stdout, NULL, _IONBF, BUFSIZ);
-
-    fprintf(stderr, "This example node shows a parameter callback that rejects\n");
-    fprintf(stderr, "all parameter updates except for those that set an even integer.\n");
-    fprintf(stderr, "Try running 'ros2 param set /even_parameters_node myint 2' to test it.\n");
+    RCLCPP_INFO(get_logger(), "This example node shows a parameter callback that rejects");
+    RCLCPP_INFO(get_logger(), "all parameter updates except for those that set an even integer.");
+    RCLCPP_INFO(get_logger(), "Try running 'ros2 param set /even_parameters_node myint 2' to");
+    RCLCPP_INFO(get_logger(), "successfully set a parameter.");
 
     // Declare a parameter change request callback
     // This function will enforce that only setting even integer parameters is allowed
--- a/demo_nodes_cpp/src/parameters/even_parameters_node.cpp
+++ b/demo_nodes_cpp/src/parameters/even_parameters_node.cpp
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <memory>
 #include <vector>
 
 #include "rclcpp/rclcpp.hpp"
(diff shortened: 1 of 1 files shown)


demo_nodes_cpp/src/parameters/even_parameters_node.cpp
```diff
@@ -31,6 +30,11 @@ class EvenParameterNode : public rclcpp::Node
   {
     // Force flush of the stdout buffer.
     setvbuf(stdout, NULL, _IONBF, BUFSIZ);
+
+    fprintf(stderr, "This example node shows a parameter callback that rejects\n");
+    fprintf(stderr, "all parameter updates except for those that set an even integer.\n");
+    fprintf(stderr, "Try running 'ros2 param set /even_parameters_node myint 2' to test it.\n");
```
wjwwood, line 36: Should this be on `stderr`? Seems like "information" to me (rather than a "warning" or "error") so why not just `RCLCPP_INFO`?

clalancette: You are totally correct.  Fixed in b7c8d5d

wjwwood (approved):

demo_nodes_cpp/src/parameters/even_parameters_node.cpp
```diff
@@ -29,8 +28,11 @@ class EvenParameterNode : public rclcpp::Node
   explicit EvenParameterNode(rclcpp::NodeOptions options)
   : Node("even_parameters_node", options.allow_undeclared_parameters(true))
   {
-    // Force flush of the stdout buffer.
-    setvbuf(stdout, NULL, _IONBF, BUFSIZ);
```
jacobperron, line 33: Just wanted to check. I know this was here to make sure tests pass. But I also vaguely recall that we've switched to unbuffered logging by default. Is that right?

demo_nodes_cpp/src/parameters/even_parameters_node.cpp
```diff
@@ -70,6 +73,7 @@ class EvenParameterNode : public rclcpp::Node
               "requested value for parameter '%s' is not an even number, rejecting change...",
               parameter.get_name().c_str()
             );
+            result.reason = "only integer parameters can be set";
```
jacobperron, line 76: Can we actually get to this case now that parameter types are static and enforced by rclcpp?

jacobperron (approved):
LGTM, just a couple questions that maybe could be addressed independently of this PR.

demo_nodes_cpp/src/parameters/even_parameters_node.cpp
clalancette: Yeah, that's correct.  In Foxy, we changed to printing logging to stderr by default.  Leaving this line in wouldn't hurt anything, it just isn't necessary anymore.

demo_nodes_cpp/src/parameters/even_parameters_node.cpp
clalancette: We can in this particular case because this node has `allow_undeclared_parameters` set to true.  Thus it acts like a parameter blackboard.  Here's a session from me testing this out:

```
$ ros2 param set /even_parameters_node myint 2
Set parameter successful
$ ros2 param set /even_parameters_node myint 1
Setting parameter failed: only even integers can be set
$ ros2 param set /even_parameters_node mydouble 1.0
Setting parameter failed: only integer parameters can be set
```

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14431/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9203/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12103/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14531/)

PR #502, opened 2021-04-19
Use underscores instead of dashes in setup.cfg

ivanpauno:
To avoid a python user warning when building:

https://ci.ros2.org/view/nightly/job/nightly_linux_release/1890/consoleFull#console-section-446


clalancette:
Same comment as https://github.com/ros2/ros2cli/pull/627#issuecomment-822531461

clalancette (approved):
Never mind my previous comment; I retested this by hacking this change into a debian build locally, and it seems to work fine.  My suggestion here is that we merge this, re-release this package, and then double-check the packages once the buildfarm rebuilds.  Assuming that is happy, we can go ahead and review and merge the rest of them.

PR #506, opened 2021-04-21
Clean up lifecycle demo

flynneva:
this PR does a few things:
- separate `lifecycle` package to headers and source structure (add `include` directory)
- use `ament_auto` macros to help keep dependencies in check

These are the two big ones, and should help most users get a better grasp of how to implement `LifecycleNodes` going forward, as well as a more "standard" way of writing them.

### DIFF
--- a/lifecycle/CMakeLists.txt
+++ b/lifecycle/CMakeLists.txt
@@ -11,55 +11,28 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-find_package(ament_cmake REQUIRED)
-find_package(rclcpp_lifecycle REQUIRED)
-find_package(lifecycle_msgs REQUIRED)
-find_package(rclcpp REQUIRED)
-find_package(std_msgs REQUIRED)
+find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
 
-### demos
-add_executable(lifecycle_talker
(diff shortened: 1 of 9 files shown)


flynneva:
this passes the tests locally for rolling. not sure why this is failing in the CI. 
[stdout.log](https://github.com/ros2/demos/files/6351426/stdout.log)

lifecycle/CMakeLists.txt
```diff
@@ -11,55 +11,28 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-find_package(ament_cmake REQUIRED)
-find_package(rclcpp_lifecycle REQUIRED)
-find_package(lifecycle_msgs REQUIRED)
-find_package(rclcpp REQUIRED)
-find_package(std_msgs REQUIRED)
+find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
```
jacobperron, line 15: We don't really use this feature in the core ROS 2 packages, though I don't know the reason why.
Since we seem to be moving closer to vanilla CMake packages, I would rather see the `ament_cmake_auto` addition proposed as a separate PR to discuss separately from the refactor of the lifecycle demos code.

jacobperron (commented):
Sorry for the late review.

lifecycle/CMakeLists.txt
clalancette: Yes, agreed.  The main reason we don't use it is exactly to move towards more vanilla CMake, so I would suggest not doing this and simplifying this patch.

PR #507, opened 2021-04-21
Foxy: separate out into header files, use ament_auto

flynneva:
same as #506 but for foxy.

this PR does a few things:
- separate lifecycle package to headers and source structure (add include directory)
- use ament_auto macros to help keep dependencies in check

These are the two big ones, and should help most users get a better grasp of how to implement LifecycleNodes going forward, as well as a more "standard" way of writing them.


jacobperron:
I'd rather wait to see this kind of change merged for Rolling before we consider backporting.

PR #508, opened 2021-04-21
Cleanup the README.rst for the lifecycle demo.

### LINKED ISSUE #505: The lifecycle demo README needs updating
There are a couple references to 'beta 1'. Seems like the README needs a refresh to bring it up to date.

clalancette:

This should fix #505

### DIFF
--- a/lifecycle/README.rst
+++ b/lifecycle/README.rst
@@ -1,11 +1,11 @@
 Introduction
 ------------
 
-ROS2 introduces the concept of managed nodes, also called ``LifecycleNode``\ s.
+ROS 2 introduces the concept of managed nodes, also called ``LifecycleNode``\ s.
 In the following tutorial, we explain the purpose of these nodes, what makes them different from regular nodes and how they comply to a lifecycle management.
 Managed nodes are scoped within a state machine of a finite amount of states.
 These states can be changed by invoking a transition id which indicates the succeeding consecutive state.
-The state machine is implemented as described at the `ROS2 design page <http://design.ros2.org/articles/node_lifecycle.html>`__.
+The state machine is implemented as described at the `ROS 2 design page <http://design.ros2.org/articles/node_lifecycle.html>`__.
 
 Our implementation differentiates between ``Primary States`` and ``Transition States``.
 Primary States are supposed to be steady states in which any node can do the respected task.
@@ -15,14 +15,12 @@ Thus, any managed node can be in one of the following states:
 
 Primary States (steady states):
 
-
 * unconfigured
 * inactive
 * active
 * shutdown
 
 Transition States (intermediate states):
-
 
 * configuring
 * activating
@@ -31,7 +29,6 @@ Transition States (intermediate states):
 * shuttingdown
 
 The possible transitions to invoke are:
-
 
 * configure
 * activate
@@ -47,8 +44,7 @@ The demo
 What's happening
 ^^^^^^^^^^^^^^^^
 
-The demo is split into 3 different separate applications.
-
+The demo is split into 3 separate applications:
 
 * lifecycle_talker
 * lifecycle_listener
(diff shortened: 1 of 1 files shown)


lifecycle/README.rst
```diff
@@ -256,7 +248,7 @@ It returns a list of all possible transitions this node can execute.
 ros2 lifecycle command line interface
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The ``lifecycle_service_client`` application is a fixed order script for this demo purpose only.
+The ``lifecycle_service_client`` application is a fixed order script for demo purposes only.
```
hidmic, line 251: @clalancette I understand what it means by _fixed order_, but perhaps it can be phrased better?

clalancette, line 251:
```suggestion
The ``lifecycle_service_client`` application executes a fixed set of operations for demo purposes only.
```

clalancette:
I made the changes you suggested in 2528366 .  Please take another look.

lifecycle/README.rst
```diff
@@ -1,11 +1,11 @@
 Introduction
 ------------
 
-ROS2 introduces the concept of managed nodes, also called ``LifecycleNode``\ s.
+ROS 2 introduces the concept of managed nodes, also called ``LifecycleNode``\ s.
 In the following tutorial, we explain the purpose of these nodes, what makes them different from regular nodes and how they comply to a lifecycle management.
 Managed nodes are scoped within a state machine of a finite amount of states.
```
mjeronimo, line 6: I don't think this sentence is very clear. Managed nodes implement a state machine. I don't know what "scoped within a state machine" is referring to and I think that a finite number of states goes without saying.

lifecycle/README.rst
```diff
@@ -1,11 +1,11 @@
...
 These states can be changed by invoking a transition id which indicates the succeeding consecutive state.
-The state machine is implemented as described at the `ROS2 design page <http://design.ros2.org/articles/node_lifecycle.html>`__.
+The state machine is implemented as described at the `ROS 2 design page <http://design.ros2.org/articles/node_lifecycle.html>`__.
 
 Our implementation differentiates between ``Primary States`` and ``Transition States``.
 Primary States are supposed to be steady states in which any node can do the respected task.
```
mjeronimo, line 11: I think this terminology is awkward. A state machine has states and transitions. I think it's potentially confusing to define a "transition state."

hidmic (approved):

lifecycle/README.rst
```diff
@@ -63,19 +59,19 @@ We split the tasks of the talker node into separate pieces and execute them as f
 #. cleanup: We destroy the publisher and timer
 
 The principle is implemented in this demo as the typical talker/listener demo.
```
mjeronimo, line 61: What principle? Instead of mentioning a principle that wasn't introduced could simply say that the demo implements a lifecycle node.

lifecycle/README.rst
```diff
@@ -63,19 +59,19 @@ We split the tasks of the talker node into separate pieces and execute them as f
...
-However, imaging a real scenario with attached hardware which may have a rather long booting phase, i.e. a laser or camera.
-One could image bringing up the device driver in the configuring state, start and stop only the publishing of the device's data and only in the cleanup/shutdown phase actually shutdown the device.
+However, imagine a real scenario with attached hardware which may have a rather long booting phase, i.e. a laser or camera.
```
mjeronimo, line 62: "However" would indicate some kind of contrast, but there isn't one. It's superfluous here and it could just say, "Imagine a real..."

lifecycle/README.rst
```diff
@@ -63,19 +59,19 @@ We split the tasks of the talker node into separate pieces and execute them as f
...
+One could imagine bringing up the device driver in the configuring state, start and stop only the publishing of the device's data in active/deactive state, and only in the cleanup/shutdown state actually shutdown the device.
 
 The ``lifecycle_listener`` is a simple listener which shows the characteristics of the lifecycle talker.
-The talker enables the message publishing only in the active state and thus making the listener receiving only messages when the talker is in an active state.
+The ``lifecycle_talker`` enables message publishing only in the active state and thus the listener only receives messages when the talker is in an active state.
```
mjeronimo, line 66: An aside: unfortunately, subscribers are not lifecycle enabled. This creates a bit of an issue because a lifecycle node that is not yet in the Active state can receive messages and so has to implement its own mechanism to avoid receiving messages when not in the Active state. This can happen when lifecycle nodes are mixed with other nodes (publishers) that are not lifecycle enabled.

lifecycle/README.rst
clalancette: While I agree with you in an abstract sense, we do indeed define "transition states" as part of the state machine: https://github.com/ros2/rcl_interfaces/blob/master/lifecycle_msgs/msg/State.msg .  So this sentence matches the code, even though the terminology is a bit weird.

lifecycle/README.rst
clalancette: Agreed.  That problem is tracked somewhere else (I can't find the ticket at the moment), but I'm not sure we should update anything here in the documentation about that.  We'll obviously have to update this once we implement a subscriber state machine.

lifecycle/README.rst
clalancette: I didn't make any changes here, as with the previous cleanup I think this is fairly clear now.

lifecycle/README.rst
clalancette: I refactored it in 3556153

lifecycle/README.rst
mjeronimo: Ok. Good that the docs are in sync.

mjeronimo (approved):

clalancette:
Merging without CI as this is only in the README file.  Thanks for the reviews!

clalancette:
@Mergifyio backport galactic

PR #515, opened 2021-04-29
Use sizeof(char) in place for sizeof(void)

hidmic:
The latter is invalid C++, even though gcc accepts it as an extension. Connected to https://github.com/ros2/rclcpp/pull/1647.

CI up to `rclcpp` and `demo_nodes_cpp`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14524/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9281/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12190/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14630/)

### DIFF
--- a/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
+++ b/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
@@ -16,6 +16,7 @@
 #include <list>
 #include <memory>
 #include <string>
+#include <type_traits>
 #include <utility>
 
 #include "rclcpp/rclcpp.hpp"
@@ -56,7 +57,10 @@ public:
       return nullptr;
     }
     num_allocs++;
-    return static_cast<T *>(std::malloc(size * sizeof(T)));
+    // Use sizeof(char) in place for sizeof(void)
+    constexpr size_t value_size = sizeof(
+      typename std::conditional<!std::is_void<T>::value, T, char>::type);
+    return static_cast<T *>(std::malloc(size * value_size));
   }
 
   void deallocate(T * ptr, size_t size)


wjwwood (approved):

clalancette:
> The latter is invalid C++, even though gcc accepts it as an extension. Connected to [ros2/rclcpp#1647](https://github.com/ros2/rclcpp/pull/1647).

I'm not against making this change, but this can't just be a GCC extension, right?  This same code is compiling with clang and MSVC as well.

wjwwood:
Yeah, I dunno, `libcxx` (which works on windows afaik) just straight does it:

https://github.com/llvm/llvm-project/blob/5bf2ef9d869ba882480232a4ed87728af74dac3b/libcxx/include/__memory/allocator.h#L74-L84

There is a specialization for `void`, but it doesn't affect that method, afaict:

https://github.com/llvm/llvm-project/blob/aaf026d9da3885a951dcdc5edd64c8e7d23b6285/libcxx/include/__memory/allocator.h#L31-L53

ivanpauno:
> I'm not against making this change, but this can't just be a GCC extension, right? This same code is compiling with clang and MSVC as well.

This seems to be a broadly accepted extension :joy:.

But AFAIU it's not supposed to be correct.
According to cppreference:

> sizeof cannot be used with function types, incomplete types (including void), or bit field lvalues.

ivanpauno (approved):

hidmic:
FWIW, besides `sizeof(void)` being accepted by some compilers, I suspect templates' lazy instantiation has a role here. IIUC if the `allocate` method in an allocator bound to `void` is ever called or even addressed, its definition won't ever be instantiated (referred to as _partial instantiation_ in [C++ Templates - The Complete Guide](http://www.tmplbook.com/), section 14.2).

PR #516, opened 2021-05-05
Revert "Use sizeof(char) in place for sizeof(void) (#515)"

hidmic:
This reverts commit c3769dd063e6c16b6fe1d3f2943408a40ff58ec6.

I was wrong. `void` allocators shouldn't behave like `char` allocators. See https://github.com/ros2/rclcpp/pull/1657.

### DIFF
--- a/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
+++ b/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
@@ -16,7 +16,6 @@
 #include <list>
 #include <memory>
 #include <string>
-#include <type_traits>
 #include <utility>
 
 #include "rclcpp/rclcpp.hpp"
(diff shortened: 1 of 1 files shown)


hidmic:

hidmic:
CI up to `demo_nodes_cpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14584/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9339/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12248/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14712/)

ivanpauno (approved):

clalancette (commented):
I may not be qualified to review this, but I'm trying to reconcile a couple of different statements made around this area:

* https://github.com/ros2/rclcpp/pull/1657#issue-630596402 says "it doesn't cover custom void allocators (which shouldn't have allocate nor deallocate methods to begin with)."
* https://en.cppreference.com/w/cpp/memory/allocator says "The explicit specialization for void lacks the member typedefs reference, const_reference, size_type and difference_type. This specialization declares no member functions. " (which agrees with the above statement)
* This file (allocator_tutorial.cpp) declares `struct MyAllocator`, which by default has `typename T = void`, and also has implementations of `allocate` and `deallocate`.  Doesn't that directly contradict the above two statements?

hidmic:
Alright, thanks for the reviews ! Going in.

PR #523, opened 2021-06-04
add NoDL files for talker and listener

wjwwood:
I'm using this to test some things where we combine `ros2 launch` and SROS2.


demo_nodes_cpp/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(ament_nodl REQUIRED)
```
hidmic, line 15: @wjwwood meta: I think we should make this an optional dependency. Core builds (that include `demo_nodes_cpp` but not `ament_nodl`) will fail otherwise.

PR #524, opened 2021-06-16
Updated deprecated use of scripts_dir and install_scripts

zmk5:
Compiling this package in ROS 2 Galactic on macOS with Python 3.9 yields a deprecation warning about the use of `scripts_dir` and `install_scripts`. For newer versions of Python and `setuptools`, the correct usage is `scripts-dir` and `install-scripts`.


zmk5:
Apologizes everyone! I mixed the two. The *underscore* is the correct one and the "dash" is the deprecated versions!

PR #526, opened 2021-07-05
Add loaned message listener.

llapx:
Signed-off-by: Lei Liu <Lei.Liu.AP@sony.com>

### DIFF
--- a/demo_nodes_cpp/src/topics/listener_loaned_message.cpp
+++ b/demo_nodes_cpp/src/topics/listener_loaned_message.cpp
@@ -1,0 +1,53 @@
+// Copyright 2014 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp_components/register_node_macro.hpp"
+
+#include "std_msgs/msg/float64.hpp"
+#include "demo_nodes_cpp/visibility_control.h"
+
+namespace demo_nodes_cpp
+{
+  // Create a Listener class that subclasses the generic rclcpp::Node base class.
+  // The main function below will instantiate the class as a ROS node.
+  class LoanedMessageListener : public rclcpp::Node
+  {
+  public:
+    DEMO_NODES_CPP_PUBLIC
+    explicit LoanedMessageListener(const rclcpp::NodeOptions &options)
+        : Node("listener_loaned_message", options)
+    {
+      // Create a callback function for when messages are received.
+      // Variations of this function also exist using, for example UniquePtr for zero-copy transport.
+      setvbuf(stdout, NULL, _IONBF, BUFSIZ);
+      auto callback =
+          [this](const std_msgs::msg::Float64::SharedPtr msg) -> void
+      {
+        RCLCPP_INFO(this->get_logger(), "I heard: [%f]", msg->data);
+      };
+      // Create a subscription to the topic which can be matched with one or more compatible ROS
+      // publishers.
+      // Note that not all publishers on the same topic with the same type will be compatible:
+      // they must have compatible Quality of Service policies.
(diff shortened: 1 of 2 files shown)


llapx:
During my test for zero copy, I found there's only talker_loaned_message, so I add a pair for it, hope it's useful.
Thanks.

Karsten1987:
isn't that an exact duplicate of the listener demo? While I am all up for consistency, I am not really a fan of copying code.

llapx:
> isn't that an exact duplicate of the listener demo? While I am all up for consistency, I am not really a fan of copying code.

Thanks for your reply!
listener.cpp only accept type `String`, can not be used for the case of loaned message, so I add it which paired with talker_loaned_message.cpp.

demo_nodes_cpp/src/topics/listener_loaned_message.cpp
```diff
@@ -0,0 +1,53 @@
...
+class LoanedMessageListener : public rclcpp::Node
+{
+public:
+  DEMO_NODES_CPP_PUBLIC
+  explicit LoanedMessageListener(const rclcpp::NodeOptions & options)
+  : Node("listener_loaned_message", options)
+  {
+    // Create a callback function for when messages are received.
+    // Variations of this function also exist using, for example UniquePtr for zero-copy transport.
+    setvbuf(stdout, NULL, _IONBF, BUFSIZ);
```
clalancette, line 34: I know we have this in our other examples, but it actually isn't necessary anymore.  Since Foxy, we print out all `RCLCPP_*` output to stderr.  So just remove this.

demo_nodes_cpp/src/topics/listener_loaned_message.cpp
```diff
@@ -0,0 +1,53 @@
...
+    auto callback =
+      [this](const std_msgs::msg::Float64::SharedPtr msg) -> void
```
clalancette, line 36: I think it is import to have a comment that shows that this can only be zero-copy for fixed-sized types.

demo_nodes_cpp/src/topics/listener_loaned_message.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "rclcpp_components/register_node_macro.hpp"
+
+#include "std_msgs/msg/float64.hpp"
+#include "demo_nodes_cpp/visibility_control.h"
+
+namespace demo_nodes_cpp
+{
+// Create a Listener class that subclasses the generic rclcpp::Node base class.
+// The main function below will instantiate the class as a ROS node.
+class LoanedMessageListener : public rclcpp::Node
```
clalancette, line 25: Personally, I would rename this class.  The fact that it is a loaned message under the hood isn't super-interesting to the user.  What's more interesting is that it is zero-copy, so I'd rename this to `ZeroCopyListener` or whatever.

clalancette (changes requested):
Besides the inline comments, I think it is important to point out what the expectation of this example is.  That is, in what circumstances should the user expect this to be zero-copy?  How would they verify this is the case?

demo_nodes_cpp/src/topics/listener_loaned_message.cpp
Karsten1987: I kind of disagree with this. We have the equivalent talker which is called `LoanedMessageTalker` and I believe this pattern makes sense, nonetheless for consistency - we also have a pair of `SerializedMessage[Talker|Listener]`.

The idiom of a loaned message is needed for zero-copy, but you can perfectly fine use a loaned message without the middleware supporting zero-copy.

demo_nodes_cpp/src/topics/listener_loaned_message.cpp
Karsten1987: You could print some information about the listener in the constructor of this class and print whether the middleware supports zero-copy.
We could enhance the talker class by printing the pointer address of the message to be published and print the pointer address of the message you get here in the callback. In the case of zero-copy these addresses are the same.

That way we can demonstrate both, a loaned message where the middleware allocates all messages and a zero-copy transport in case the middleware supports it.

Karsten1987:
> listener.cpp only accept type String, can not be used for the case of loaned message, so I add it which can be paired with talker_loaned_message.cpp.

This makes total sense. Thanks for the clarification.

demo_nodes_cpp/src/topics/listener_loaned_message.cpp
clalancette: > I kind of disagree with this. We have the equivalent talker which is called `LoanedMessageTalker` and I believe this pattern makes sense, nonetheless for consistency - we also have a pair of `SerializedMessage[Talker|Listener]`.
> 
> The idiom of a loaned message is needed for zero-copy, but you can perfectly fine use a loaned message without the middleware supporting zero-copy.

OK, that makes sense.  I think the thing I'm stuck on here is how the user knows that loaning is happening, and what they should expect different when loaning is used.

PR #539, opened 2021-11-08
Update python nodes SIGINT handling

ivanpauno:
Related to https://github.com/ros2/rclpy/pull/844.

### DIFF
--- a/demo_nodes_py/demo_nodes_py/topics/listener.py
+++ b/demo_nodes_py/demo_nodes_py/topics/listener.py
@@ -12,7 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 import rclpy
+from rclpy.executor import ExternalShutdownException
 from rclpy.node import Node
 
 from std_msgs.msg import String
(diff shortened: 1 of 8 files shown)


hidmic (commented):
Same question as in https://github.com/ros2/system_tests/pull/490#pullrequestreview-801280689.

hidmic (approved):
LGTM pending green CI

demo_nodes_py/demo_nodes_py/topics/listener.py
```diff
@@ -36,9 +39,11 @@ def main(args=None):
         rclpy.spin(node)
     except KeyboardInterrupt:
         pass
-
-    node.destroy_node()
-    rclpy.shutdown()
+    except ExternalShutdownException:
+        sys.exit(1)
```
jacobperron, line 43: Sorry to comment on an old ticket, but why are we exiting with error code 1? IIUC, `ExternalShutdownException` is something we should expect as part of the normal shutdown sequence here.

It seems to me that all of the demos should be updated to look like,

```py
try:
    rclpy.spin(node)
except (KeyboardInterrupt, ExternalShutdownException):
    pass
```

jacobperron: See https://github.com/ros2/demos/pull/581

Issue #540, opened 2021-11-09
vexing parse problem leads to compile error


fujitatomoya:
This is not a feature request nor bug fix. Just an improvement for demo sample code to address possible vexing parse problem that user might meet.

## How to produce the problem

the following cannot be compiled since vexing parse problem.

```diff
diff --git a/demo_nodes_cpp/src/topics/talker.cpp b/demo_nodes_cpp/src/topics/talker.cpp
index 30f97a0..8cdba3c 100644
--- a/demo_nodes_cpp/src/topics/talker.cpp
+++ b/demo_nodes_cpp/src/topics/talker.cpp
@@ -50,7 +50,7 @@ public:
         pub_->publish(std::move(msg_));
       };
     // Create a publisher with a custom Quality of Service profile.
-    rclcpp::QoS qos(rclcpp::KeepLast(7));
+    rclcpp::QoS qos(rclcpp::KeepAll());
     pub_ = this->create_publisher<std_msgs::msg::String>("chatter", qos);

     // Use a timer to schedule periodic message publishing.
```

compile error is

```
root@tomoyafujita:~/ros2_ws/colcon_ws# colcon build --symlink-install --packages-select demo_nodes_cpp
Starting >>> demo_nodes_cpp
--- stderr: demo_nodes_cpp
/root/ros2_ws/colcon_ws/src/ros2/demos/demo_nodes_cpp/src/topics/talker.cpp: In constructor ‘demo_nodes_cpp::Talker::Talker(const rclcpp::NodeOptions&)’:
/root/ros2_ws/colcon_ws/src/ros2/demos/demo_nodes_cpp/src/topics/talker.cpp:54:72: error: no matching function for call to ‘demo_nodes_cpp::Talker::create_publisher<std_msgs::msg::String>(const char [8], rclcpp::QoS (&)(rclcpp::KeepAll (*)()))’
   54 |     pub_ = this->create_publisher<std_msgs::msg::String>("chatter", qos);
      |                                                                        ^
In file included from /root/ros2_ws/colcon_ws/install/rclcpp/include/rclcpp/executors/single_threaded_executor.hpp:28,
                 from /root/ros2_ws/colcon_ws/install/rclcpp/include/rclcpp/executors.hpp:22,
                 from /root/ros2_ws/colcon_ws/install/rclcpp/include/rclcpp/rclcpp.hpp:155,
[...]

PR #541, opened 2021-11-09
Add how to fix the most vexing parse problem

fujitatomoya:
address https://github.com/ros2/demos/issues/540

Signed-off-by: Tomoya Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/demo_nodes_cpp/src/topics/talker.cpp
+++ b/demo_nodes_cpp/src/topics/talker.cpp
@@ -52,7 +52,7 @@ public:
     // Create a publisher with a custom Quality of Service profile.
     // Extra pair of parentheses to address vexing parse problem.
     // Or use uniform initialization.
-    rclcpp::QoS qos((rclcpp::KeepLast(7)));
+    rclcpp::QoS qos(rclcpp::KeepLast{7});
     pub_ = this->create_publisher<std_msgs::msg::String>("chatter", qos);
 
     // Use a timer to schedule periodic message publishing.
--- a/demo_nodes_cpp/src/topics/talker.cpp
+++ b/demo_nodes_cpp/src/topics/talker.cpp
@@ -50,8 +50,9 @@ public:
         pub_->publish(std::move(msg_));
       };
     // Create a publisher with a custom Quality of Service profile.
-    // Extra pair of parentheses to address vexing parse problem.
-    // Or use uniform initialization.
+    // Uniform initialization is suggested so it can be trivially changed to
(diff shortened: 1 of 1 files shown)


clalancette:
Can I suggest uniform initialization instead?  It's more straightforward, and we required C++14 (for Foxy) or C++17 (for Galactic and Rolling).

I'd still keep the comment, though updated to explain why uniform initialization must be used here.

fujitatomoya:
@clalancette sounds good to me, thanks for checking!

demo_nodes_cpp/src/topics/talker.cpp
```diff
@@ -50,7 +50,9 @@ class Talker : public rclcpp::Node
         pub_->publish(std::move(msg_));
       };
     // Create a publisher with a custom Quality of Service profile.
-    rclcpp::QoS qos(rclcpp::KeepLast(7));
+    // Extra pair of parentheses to address vexing parse problem.
+    // Or use uniform initialization.
```
clalancette, line 54:
```suggestion
    // Uniform initialization is suggested so it can be trivially changed to rclcpp::KeepAll{} if the user wishes.
    // (rclcpp::KeepLast(7) -> rclcpp::KeepAll() fails to compile)
```

clalancette (changes requested):
One more small item to fix in the comments, then I think this is good to go.

That said, I'm curious why `rclcpp::QoS qos(rclcpp::KeepAll());` fails to compile.  Naively, I expect it would.  Should we file a bug in `rclcpp` to follow-up and fix that?

demo_nodes_cpp/src/topics/talker.cpp
fujitatomoya: > why rclcpp::QoS qos(rclcpp::KeepAll()); fails to compile.

i was there too, as far as i know for now, it is about C++ compiler parse. (i am not sure if this is a problem, but it really bothers me and took time to check.) So i just thought it would be better to file this to avoid possible future problem.

clalancette (approved):
Fine by me with green CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15583/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10267/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13251/)
* Windows [Build Status]

fujitatomoya:
all green, going in.

PR #542, opened 2021-11-12
Fixed typo executor -> executors

ori155:
quick fix of a typo which prevented the `listener` node from running.

### DIFF
--- a/demo_nodes_py/demo_nodes_py/topics/listener.py
+++ b/demo_nodes_py/demo_nodes_py/topics/listener.py
@@ -15,7 +15,7 @@
 import sys
 
 import rclpy
-from rclpy.executor import ExternalShutdownException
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Good catch, I'm surprised our CI didn't catch this.  @ivanpauno can you take a look and run CI here?

ivanpauno:
> Good catch, I'm surprised our CI didn't catch this. @ivanpauno can you take a look and run CI here?

I guess there's no test for this ...

ivanpauno (approved):
Thanks for the fix @ori155 !!

ivanpauno:
> @ivanpauno can you take a look and run CI here?

The Rpr checker is passing, considering there's no test for this executable I'm not going to run CI.
I did test locally, this fixes the issue.

PR #551, opened 2022-01-07
Fix include order and relative paths for cpplint

jacobperron:
Relates to https://github.com/ament/ament_lint/pull/324

### DIFF
--- a/intra_process_demo/src/image_pipeline/camera_node.cpp
+++ b/intra_process_demo/src/image_pipeline/camera_node.cpp
@@ -16,7 +16,7 @@
 
 #include "image_pipeline/camera_node.hpp"
 
-#include <rclcpp/rclcpp.hpp>
+#include "rclcpp/rclcpp.hpp"
 
 int main(int argc, char ** argv)
 {
--- a/intra_process_demo/src/image_pipeline/camera_node.cpp
+++ b/intra_process_demo/src/image_pipeline/camera_node.cpp
@@ -12,11 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
+
 #include "image_pipeline/camera_node.hpp"
 
 #include <rclcpp/rclcpp.hpp>
-
-#include <memory>
 
 int main(int argc, char ** argv)
 {
--- a/intra_process_demo/src/image_pipeline/image_pipeline_all_in_one.cpp
+++ b/intra_process_demo/src/image_pipeline/image_pipeline_all_in_one.cpp
@@ -14,7 +14,7 @@
 
 #include <memory>
 
-#include <rclcpp/rclcpp.hpp>
+#include "rclcpp/rclcpp.hpp"
 
 #include "image_pipeline/camera_node.hpp"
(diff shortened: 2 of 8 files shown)


intra_process_demo/src/image_pipeline/camera_node.cpp
```diff
@@ -12,12 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <memory>
+
 #include "image_pipeline/camera_node.hpp"
 
 #include <rclcpp/rclcpp.hpp>
```
clalancette, line 19: Should we change to double-quotes here?

intra_process_demo/src/image_pipeline/image_pipeline_all_in_one.cpp
```diff
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <rclcpp/rclcpp.hpp>
-
 #include <memory>
 
+#include <rclcpp/rclcpp.hpp>
```
clalancette, line 17: Same question.

intra_process_demo/src/image_pipeline/image_pipeline_with_two_image_view.cpp
```diff
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <rclcpp/rclcpp.hpp>
-
 #include <memory>
 
+#include <rclcpp/rclcpp.hpp>
```
clalancette, line 17: Same question.

intra_process_demo/src/image_pipeline/camera_node.cpp
jacobperron: It's not necessary to satisfy cpplint. Basically, if the header has the suffix ".h", then we need to use double-quotes (or move it to the top of the file) since cpplint assumes it is a C system header.

It some places where there was a mix of headers with the suffix ".h" and ".hpp", I used double-quotes everywhere for aesthetics :shrug:

clalancette: All right.  I actually don't have a strong position here, I was just going for consistency.  I'll approve as-is.

clalancette (approved):

intra_process_demo/src/image_pipeline/camera_node.cpp
jacobperron: Switched to double-quotes for backwards-compatibility with the linter (1349899).

intra_process_demo/src/image_pipeline/image_pipeline_with_two_image_view.cpp
jacobperron: 1349899

intra_process_demo/src/image_pipeline/image_pipeline_all_in_one.cpp
jacobperron: 1349899

audrow (approved):

PR #552, opened 2022-01-13
Use default on_activate()/on_deactivate implemenetation of Node

ivanpauno:
Update after https://github.com/ros2/rclcpp/pull/1863/files.

### DIFF
--- a/lifecycle/src/lifecycle_talker.cpp
+++ b/lifecycle/src/lifecycle_talker.cpp
@@ -144,12 +144,13 @@ public:
    * TRANSITION_CALLBACK_ERROR or any uncaught exceptions to "errorprocessing"
    */
   rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
-  on_activate(const rclcpp_lifecycle::State &)
+  on_activate(const rclcpp_lifecycle::State & state)
   {
-    // We explicitly activate the lifecycle publisher.
-    // Starting from this point, all messages are no longer
-    // ignored but sent into the network.
-    pub_->on_activate();
+    // The parent class method automatically transition on managed entities
+    // (currently, LifecyclePublisher).
+    // pub_->on_activate() could also be called manually here.
+    // Overriding this method is optional, a lot of times the default is enough.
+    LifecycleNode::on_activate(state);
 
     RCUTILS_LOG_INFO_NAMED(get_name(), "on_activate() is called.");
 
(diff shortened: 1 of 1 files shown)


lifecycle/src/lifecycle_talker.cpp
```diff
@@ -144,12 +144,13 @@ class LifecycleTalker : public rclcpp_lifecycle::LifecycleNode
...
   {
-    // We explicitly activate the lifecycle publisher.
-    // Starting from this point, all messages are no longer
-    // ignored but sent into the network.
-    pub_->on_activate();
+    // The parent class method automatically transition on managed entities
+    // (currently, LifecyclePublisher).
+    // pub_->on_activate() could also be called manually here.
+    // Overriding this method is optional, a lot of times the default is enough.
+    LifecycleNode::on_activate(state);
```
jacobperron, line 153: Following my comment here https://github.com/ros2/rclcpp/pull/1863#discussion_r784148928

What if we always called the parent logic automatically _after_ any user-defined callback, rather than explicitly calling the parent callback here. If desired, we could provide an option to disable automatic transitions of the managed entities.

ivanpauno: > What if we always called the parent logic automatically after any user-defined callback, rather than explicitly calling the parent callback here. If desired, we could provide an option to disable automatic transitions of the managed entities.

I don't see much value on that.
Most of the time you're not going to override on_activate()/on_deactivate().

If you override and want to keep doing what the base class is doing, you have to explicitly call the base class method.
That's what I usually expect to do when writing c++.

>  If desired, we could provide an option to disable automatic transitions of the managed entities.

It's also good IMO that it provides an easy way to ignore the default behavior.

ivanpauno: In this way we also avoid breaking API, as also existing nodes that were implementing on_activate()/on_deactivate() before don't need to be updated.

e.g. we can merge https://github.com/ros2/rclcpp/pull/1863 without merging this PR at the same time

jacobperron: > Most of the time you're not going to override on_activate()/on_deactivate().

I don't really know how LifecycleNode is used in practice, but you may be right.

The argument for avoiding an API break at this juncture sounds good to me :+1:

ivanpauno: > I don't really know how LifecycleNode is used in practice, but you may be right.

As far as I've seen, not much more that forwarding the on_activate() call to lifecycle publishers, e.g. https://github.com/ros-planning/navigation2/blob/30b405c58e6d53ba8c96381416bc4679d35a1483/nav2_dwb_controller/dwb_core/src/publisher.cpp#L121-L126.

jacobperron (approved):

ivanpauno:
CI passed [here](https://github.com/ros2/rclcpp/pull/1863#issuecomment-1013310557).
I will do a new rclcpp release before merging this.

ivanpauno:

ivanpauno:

ivanpauno:
I couldn't reproduce the issue locally, once again:

@ros-pull-request-builder retest this please.

clalancette:
@ros-pull-request-builder retest this please.

PR #558, opened 2022-04-07
Make lifecycle demo automatically exit when done

### LINKED ISSUE #504: [lifecycle demo] Launch script doesn't exit
## Bug report

- Operating System: Ubuntu 20.04
- Installation type: Debian-based installation of ROS 2
- DDS implementation: any

#### Steps to reproduce issue

```
ros2 launch lifecycle lifecycle_demo.launch.py
```

#### Expected behavior
The script exits cleanly after the demo is done. 

#### Actual behavior
The lifecycle_service_client runs through a script and exits, but the talker and listener are still active. It seems like the launch script could detect the exit of the lifecycle_service_client and exit the launch script.

sloretz:
Fixes #504

This makes the lifecyle demo automatically exit when it's completed. First this PR fixes a bug in `lifecycle_service_client` where it's stuck waiting for the future to complete forever because of ros2/rclcpp#1916. Next this PR adds a `Shutdown` action to the `on_exit` argument of the `lifecycle_service_client` `Node` so that when it shuts down the other nodes are killed too.

### DIFF
--- a/lifecycle/src/lifecycle_service_client.cpp
+++ b/lifecycle/src/lifecycle_service_client.cpp
@@ -303,6 +303,15 @@ callee_script(std::shared_ptr<LifecycleServiceClient> lc_client)
   }
 }
 
+void
+wake_executor(std::shared_future<void> future, rclcpp::executors::SingleThreadedExecutor & exec)
+{
+  future.wait();
+  // Wake the executor when the script is done
+  // https://github.com/ros2/rclcpp/issues/1916
+  exec.cancel();
+}
+
 int main(int argc, char ** argv)
 {
   // force flush of the stdout buffer.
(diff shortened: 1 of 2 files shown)


sloretz:
CI (build: `--packages-up-to lifecycle` test: `--packages-select lifecycle`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16547/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11133/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16921/)

mjcarroll (approved):

lifecycle/src/lifecycle_service_client.cpp
```diff
@@ -321,6 +330,10 @@ int main(int argc, char ** argv)
   std::shared_future<void> script = std::async(
     std::launch::async,
     std::bind(callee_script, lc_client));
+  auto wake_exec = std::async(
+    std::launch::async,
+    std::bind(wake_executor, script, std::ref(exe)));
```
wjwwood, line 335: I would say this is unnecessarily complicated. I would instead move the `callee_script` into a one-off timer with zero duration.

Or better yet, run the executor's spin in a separate thread (use `std::async` if you like) and do the `callee_script` here in the main thread, then when that function is done cancel the executor and "join" the thread it was using to spin in.

PR #561, opened 2022-05-06
Improve QoS demo documentation

jacobperron:
Try to clarify what each demo is doing and how to see usage.
Avoid repeating commands for both C++ and Python (since the usage is identical).

Much of these changes are subjective, so feel free to push back on them.

### DIFF
--- a/quality_of_service_demo/README.md
+++ b/quality_of_service_demo/README.md
@@ -1,7 +1,10 @@
 # Quality of Service Demos
 
-The demo applications in this package show the various Quality of Service settings available in ROS2.
+The demos in this package show the various Quality of Service settings available in ROS 2.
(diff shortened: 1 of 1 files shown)


audrow (approved):
These look like good changes to me!

clalancette (commented):
Some of these changes are good overall.  However, I'm slightly concerned that removing the duplication between C++ and Python will make it easier to miss testing one or the other during testing time (I've definitely found bugs in one but not the other in the past).  So I would suggest not getting rid of the duplication.

jacobperron:
I've added back the C++ example commands ([61291b9](https://github.com/ros2/demos/pull/561/commits/61291b9cc388779e08bbbfd46c97cd36e91acffd)).

PR #564, opened 2022-05-17
Fix two small bugs in the fibonacci C++ tutorial.

clalancette:
The first bug has to do with the action client, and what happens
if the server rejects the goal.  In the current code, the
client would just hang around forever.  If the server rejects
the goal, make sure the client shuts down.

The second bug has to do with the size that the action server
can support.  The code has more information, but essentially
because of the Fibonacci action message, we can only support
sequences up to order 46; anything larger than that will cause
integer wrapping.


audrow (approved):
Looks good to me!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16728/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11310/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17123/)

PR #579, opened 2022-08-01
common content filter support in the rcl

iuhilnehc-ynos:
related to https://github.com/ros2/common_content_filter/pull/1

Signed-off-by: Chen Lihui <lihui.chen@sony.com>


mjeronimo (approved):

wjwwood:
From the waffle triage meeting, @fujitatomoya do you know what the status of this effort is? Is it something we should look at for Lyrical or should we leave it as-is for now or something else?

fujitatomoya:
at the moment we agreed to develop the fallback content filtering (for when rmw implementation does not support content filtering subscription), but that is more than 4 years ago. we (Sony) were asking for the review feedback but there does not seem to be much interest from community on this fallback functionality in the rcl. for now, i will go ahead to close all the related issues and PRs about fallback filtering.

fujitatomoya:
see https://github.com/ros2/demos/pull/579#issuecomment-3757278811

PR #580, opened 2022-08-02
Remove action_msgs dependency

jacobperron:
It is now properly included in the rosidl dependency chain.

Depends on https://github.com/ros2/rosidl_defaults/pull/22


ivanpauno (approved):

jacobperron:
CI was run here: https://github.com/ros2/rosidl_defaults/pull/22#issuecomment-1220052092

PR #582, opened 2022-08-30
pendulum executor ROS2 refactorization

paulagb8894:


clalancette:
Thanks for the contribution.  Can you do a few things please:

1.  Please retarget this PR to the `rolling` branch.  We do all of our development on the `rolling` branch, then backport as needed to the stable branches.
2. Can you please provide a description of why you want to do this change, and what it adds?
3. You'll need to sign off your commit with a [DCO](https://elinux.org/Developer_Certificate_Of_Origin), otherwise the bot will be unhappy.
4. You'll need to fix the trailing whitespace changes, otherwise CI will be unhappy.

Once all of those things are fixed, and we can understand what you are trying to do here, we can consider it for inclusion.  Thank you.

PR #594, opened 2023-01-05
Update the demos to C++17.

clalancette:
The main reason to do this is so that we can compile the demos with the clang static analyzer.  As of clang++-14 (what is in Ubuntu 22.04), the default still seems to be C++14, so we need to specify C++17 so that new things in the rclcpp headers work properly.

Further, due to reasons I don't fully understand, I needed to set CMAKE_CXX_STANDARD_REQUIRED in order for clang to really use that version.  So set this as well.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17889/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12457/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18563/)

Issue #531, opened 2021-08-10
Add README


lorepieri8:
It would be good to have a README file for each demo that briefly explains what's inside.

audrow:
I think this is a good idea.

cardboardcode:
+1 from me as well. I should be available to take a look at this within the month of July 2022 if no one else is doing it. 

@lorepieri8 @audrow 

Will update accordingly here in this thread regarding progress.

cardboardcode:
## TODO
For each demo to have its own `README.md`, a total of **18** `README.md` will need to be created for the following folders.

Updated as of `1st July 2022`:

- [ ] root `README.md`
- [ ] [action_tutorials](https://github.com/ros2/demos/tree/rolling/action_tutorials)/action_tutorials_cpp
- [ ] [action_tutorials](https://github.com/ros2/demos/tree/rolling/action_tutorials)/action_tutorials_interfaces/
- [ ] [action_tutorials](https://github.com/ros2/demos/tree/rolling/action_tutorials)/action_tutorials_py
- [ ] [demos](https://github.com/ros2/demos)/composition/
- [ ] [demos](https://github.com/ros2/demos)/demo_nodes_cpp/
- [ ] [demos](https://github.com/ros2/demos)/demo_nodes_cpp_native/
- [ ] [demos](https://github.com/ros2/demos)/demo_nodes_py/
- [ ] [dummy_robot](https://github.com/ros2/demos/tree/rolling/dummy_robot)/dummy_map_server/
- [ ] [dummy_robot](https://github.com/ros2/demos/tree/rolling/dummy_robot)/dummy_robot_bringup/
- [ ] [dummy_robot](https://github.com/ros2/demos/tree/rolling/dummy_robot)/dummy_sensors/
- [ ] [dummy_robot](https://github.com/ros2/demos/tree/rolling/dummy_robot)/dummy_sensors/
- [ ] [demos](https://github.com/ros2/demos)/image_tools/
- [ ] [demos](https://github.com/ros2/demos)/intra_process_demo/
- [x] https://github.com/ros2/demos/tree/rolling/lifecycle
- [ ] [demos](https://github.com/ros2/demos)/lifecycle_py/
- [ ] [demos](https://github.com/ros2/demos)/logging_demo/
- [ ] [demos](https://github.com/ros2/demos)/pendulum_control/
- [ ] [demos](https://github.com/ros2/demos)/pendulum_msgs/
[...]

PR #597, opened 2023-01-24
Added README.md for demo_cpp_nodes_native

cardboardcode:
## Purpose of Pull Request

This pull request adds a `README.md` to **demo_nodes_cpp_native** as part of efforts to address the issue highlighted in https://github.com/ros2/demos/issues/531.

### DIFF
--- a/demo_nodes_cpp_native/README.md
+++ b/demo_nodes_cpp_native/README.md
@@ -1,0 +1,38 @@
+## **What Is This?**
+
+This demo ... 
+
+## **Build**
+
+```bash
+colcon build --packages-up-to demo_nodes_cpp_native
+```
+
+## **Run**
+
+```bash
+ros2 run demo_nodes_cpp_native talker
+```
+
+## **Verify**
+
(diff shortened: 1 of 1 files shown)


demo_nodes_cpp_native/README.md
```diff
@@ -0,0 +1,46 @@
+## **What Is This?**
+
+This demo provides an example of creating and running a ROS 2 node using the native **rmw_fastrtps_cpp** API directly, rather than doing so through vanilla **rclcpp**.
```
mjeronimo, line 3: I don't think this description is correct. Looking at the sample, the rmw_fastrtps_cpp API is only used to get and print pointers to the DomainParticipant and DataWriter. Otherwise, the sample uses a standard ROS 2 node and publisher

mjeronimo (changes requested):

demo_nodes_cpp_native/README.md
cardboardcode: Noted. Updated with revamped draft.

mjeronimo (approved):

PR #608, opened 2023-03-27
Fix the set_parameters_callback example program.

clalancette:
In particular, the 'on_set_parameter_callback' callback was checking for undeclared parameters, but that could not have been a problem since 'allow_undeclared_parameters' is set to false by default.

Instead, change the logic in that callback to arbitrarily reject parameters > 5.0 (for param1) or < -5.0 (for param2). That is a much better demonstration of what this node should do.

Also add in a print during the constructor, telling the user what to expect.


mjeronimo (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18318/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12840/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18984/)

PR #615, opened 2023-05-08
Use not-deprecated rclpy import.

clalancette:
qos_event was deprecated during Iron development, so switch to the non-deprecated event_handler instead.


clalancette:
This should also be backported to Iron.

mjcarroll (approved):

clalancette:
@mergifyio backport iron

PR #635, opened 2023-05-20
Add test package for single and multi thread

icsl-Jeon:
I added codes to test `SingleThreadedExecutor ` and `MultiThreadedExecutor `. 
Especially, two different callback groups can be tested.  

Although I have full descriptions and analysis in [this link](https://motion-boseong.vercel.app/threading-ros2), I could not decide whether to put them in this repo or, it is okay to have them just in a link in `README.md`.

### DIFF
--- a/threads/CMakeLists.txt
+++ b/threads/CMakeLists.txt
@@ -1,0 +1,34 @@
+cmake_minimum_required(VERSION 3.5)
+project(thread_tester)
+
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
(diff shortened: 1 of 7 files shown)


clalancette:
We already have some examples of executors in use in https://github.com/ros2/examples/tree/rolling/rclcpp/executors .  Can you take a look to see if those examples already have if you want?  If not, can you see if it makes sense to add these examples there instead?

icsl-Jeon:
Thank you for guiding me. 
I will try to open PR there 😀
Let me close this! 
Regards

PR #637, opened 2023-05-23
Cleanup the interactive quality of service demos.

### LINKED ISSUE #633: Quality of service demo interactive_{publisher,subscriber} doesn't quit properly
The quality of service interactive publisher and subscriber are run like the following:

```
ros2 run quality_of_service_demo_cpp interactive_publisher
```

or

```
ros2 run quality_of_service_demo_cpp interactive_subscriber
```

Both demos are interactive in the sense that they listen for keyboard commands to do certain actions.  One of those actions is to quit, which is done by hitting the 'x' button.

Unfortunately, this doesn't work and doesn't exit the program for either of them.  The basic problem is that the keyboard handling code that is common to both demos is doing the equivalent of a `getch` with no timeout in a thread.  So when the user hits 'x', the executor is shutdown, but nothing makes the `getch` return and thus lets the thread complete.

I'm not quite sure how to solve it at the moment, but it should be looked into.

This came out of the testing in https://github.com/osrf/ros2_test_cases/issues/783

clalancette:
The interactive demos are supposed to listen for commands on stdin and do particular actions.  And for the most part they do that, except for 'x'.  That command is supposed to quit the program, but doesn't currently do so.

What we do here is to revamp the keyboard handling so that it works in all situations.  This involves taking the executor and putting it into a thread, and then having the main thread do stdin handling.  We also install a signal handler.

With all of that in place, then, we can either catch the signal from a Ctrl-C or type 'x' to have things quit.  This then quits the keyboard handling loop, which then goes on to cancel the executor, join the thread, and shutdown rclcpp.

This fixes #633.

### DIFF
--- a/quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
+++ b/quality_of_service_demo/rclcpp/src/interactive_publisher.cpp
@@ -12,10 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <chrono>
 #include <iomanip>
 #include <iostream>
 #include <memory>
 #include <string>
+#include <thread>
 
 #include "std_msgs/msg/string.hpp"
 #include "rcutils/cmdline_parser.h"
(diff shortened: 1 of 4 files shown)


clalancette:
All right, I've finally been able to test this out on Windows.  That involved actually a rewrite of the implementation, but I think it is a lot nicer now; all of the details are hidden in `utils.cpp`.  I'm going to mark this as ready for review and run CI on it.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19283/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13821/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20008/)

audrow (approved):
LGTM!

PR #638, opened 2023-06-07
Cleanup demo_nodes_cpp CMake and dependencies

clalancette:
I was looking at demo_nodes_cpp, and I noticed a few different things:

1.  The includes in the various source files weren't always correct.  We were missing some includes, and we had some other includes that weren't actually being used.
2.  The dependencies weren't correctly specified, in that we were missing a couple.
3.  We don't really need to use `ament_target_dependencies` in the CMakeLists.txt; we can instead use `target_link_libraries`.
4.  There was actually quite a bit of duplication in the various install targets.  We could probably use a function to reduce duplication.

This PR fixes all of that, and removes about 10 lines of code in the process.

### DIFF
--- a/demo_nodes_cpp/CMakeLists.txt
+++ b/demo_nodes_cpp/CMakeLists.txt
@@ -23,18 +23,44 @@ find_package(rcutils REQUIRED)
 find_package(rmw REQUIRED)
 find_package(std_msgs REQUIRED)
 
-include_directories(include)
-
 function(custom_executable subfolder target)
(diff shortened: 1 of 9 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18869/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13409/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19623/)

ahcorde (approved):

Issue #650, opened 2023-09-28
[allocator_tutorial] Adding polymorphic allocators 


AliAshkaniNia:
I recently reviewed the [Implementing a custom memory allocator](https://docs.ros.org/en/rolling/Tutorials/Advanced/Allocator-Template-Tutorial.html) section and wanted to provide a friendly suggestion. While the `MyAllocator` implementation works well for C++11, ROS 2 also supports C++17 polymorphic memory allocators which offer a simpler alternative. I believe adding examples using the polymorphic allocators could make this section more beginner-friendly.

I am prepared to provide a minimum working example and make the necessary modifications to the documentation. Please let me know if there are any particular reasons for not including this approach in the tutorial already, or if you think it would be valuable to add as an additional option.

clalancette:
I'm not all familiar with the polymorphic allocators, but if they end up being a simpler alternative I think that would be welcome.  If you'd like to submit a PR with that change, we are happy to take a look at it.

PR #653, opened 2023-10-05
Replacing old-style C++ allocator with a polymorphic memory resource (PMR)

AliAshkaniNia:
This is a basic example mimicking the original demo. We could use a more advanced resource, like `unsynchronized_pool_resource`, for a realistic scenario.

### DIFF
--- a/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
+++ b/demo_nodes_cpp/src/topics/allocator_tutorial.cpp
@@ -15,9 +15,9 @@
 #include <chrono>
 #include <list>
 #include <memory>
+#include <memory_resource>
 #include <string>
 #include <utility>
-#include <memory_resource>
 
 #include "rclcpp/rclcpp.hpp"
 #include "rclcpp/allocator/allocator_common.hpp"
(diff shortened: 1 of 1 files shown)


AliAshkaniNia:
This is a change for issue #650  .

fujitatomoya (changes requested):
some uncrustify errors, https://build.ros2.org/job/Rpr__demos__ubuntu_jammy_amd64/235/testReport/junit/demo_nodes_cpp/uncrustify/src_topics_allocator_tutorial_cpp/

and, could you also address DCO error?

AliAshkaniNia:
My apologies, I should have read the ROS2 developer guides before making changes. Now, it is compliant with ROS2 style, and I have also reordered the `#include` statements.

fujitatomoya (approved):
> We could use a more advanced resource, like unsynchronized_pool_resource, for a realistic scenario.

I am not sure about this, we could. but we would not want to introduce the polymorphic memory details here, instead we would like it to be simple for demonstration code.

clalancette:
> I am not sure about this, we could. but we would not want to introduce the polymorphic memory details here, instead we would like it to be simple for demonstration code.

Yeah, I agree.  What we have here is easy to understand, and is the main point of this tutorial, so I think we should keep it simple.

clalancette (approved):
Thanks, this is much easier to understand.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19647/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14155/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20371/)

AliAshkaniNia:
> Thanks, this is much easier to understand.

Glad I could assist!

AliAshkaniNia:
For the `unsynchronized_pool_resource` stuff, I agree with you. I'm a fan of making tutorials dead simple, which is why I came up with this refactoring idea. Please correct me if I'm wrong, but on the other hand, my perception is that ROS2 comes with "*batteries included*"; it provides all the necessary tools and resources for creating production-ready code. I believe there's a good chance that someone interested in custom allocators might also be interested in using stack memory instead of heap. So, I thought it's worth mentioning that topic thanks to the intuitive API:
```C++
std::array<std::uint8_t, 10*1024> buffer{};
std::pmr::monotonic_buffer_resource upstream_resource(buffer.data(), buffer.size());
std::pmr::unsynchronized_pool_resource mem_resource(&upstream_resource);

auto alloc = std::make_shared<Alloc>(&mem_resource);
```

Again, this is just a suggestion, and overall, I agree that it might be too complicated for the tutorial. I just mentioned it for the sake of completeness.

PR #665, opened 2024-02-07
Update maintainer list in package.xml files

mjeronimo:
I'm no longer a maintainer and don't need to be credited as an author.


clalancette (approved):
This looks good to me.  Since this is only changing metadata in the package.xml, I'm going to merge it with just the Rpr jobs passing.

PR #680, opened 2024-05-06
Change occupancy value in `dummy_map_server` to  1.

marcoag:
As pointed out in https://github.com/osrf/ros2_test_cases/issues/1200, dummy map server occupancy value should be `1` as specified in https://github.com/ros2/common_interfaces/blob/rolling/nav_msgs/msg/OccupancyGrid.msg.

We should backport this once merged.

### DIFF
--- a/dummy_robot/dummy_map_server/src/dummy_map_server.cpp
+++ b/dummy_robot/dummy_map_server/src/dummy_map_server.cpp
@@ -65,7 +65,7 @@ int main(int argc, char * argv[])
     msg.data[(center) % (msg.info.width * msg.info.height)] = -1;
     msg.data[(rhs) % (msg.info.width * msg.info.height)] = -1;
     msg.data[(++lhs) % (msg.info.width * msg.info.height)] = 0;
-    msg.data[(++center) % (msg.info.width * msg.info.height)] = 100;
+    msg.data[(++center) % (msg.info.width * msg.info.height)] = 1;
     msg.data[(++rhs) % (msg.info.width * msg.info.height)] = 0;
 
     msg.header.stamp = clock->now();


dummy_robot/dummy_map_server/src/dummy_map_server.cpp
```diff
@@ -65,7 +65,7 @@ int main(int argc, char * argv[])
     msg.data[(center) % (msg.info.width * msg.info.height)] = -1;
     msg.data[(rhs) % (msg.info.width * msg.info.height)] = -1;
     msg.data[(++lhs) % (msg.info.width * msg.info.height)] = 0;
-    msg.data[(++center) % (msg.info.width * msg.info.height)] = 100;
+    msg.data[(++center) % (msg.info.width * msg.info.height)] = 1;
```
clalancette, line 68: I talked to @marcoag about this offline; I have some vague recollection that `100` means something in occupancy grids.

That said, I think that either this number should match what we have in the documentation (which does indeed say to use `1` here), or the documentation should be updated to specifically say that `100` has meaning.

marcoag: So in [ROS noetic](https://github.com/ros/common_msgs/blob/noetic-devel/nav_msgs/msg/OccupancyGrid.msg) it seems that it was 100 to represent percentage based probabilities. Then this was changed in the [last comments update](https://github.com/ros2/common_interfaces/pull/117).

There were some efforts to change this to 0-255 for higher resolution [here](https://github.com/ros2/common_interfaces/issues/96) but it was not pursued in order to avoid disruption to users which makes me thing a change to 0-1 would have also been avoided. It seems that now nav2 uses the 0-255 range in costmaps and translates to [0-100 range](https://github.com/open-navigation/navigation2/blob/b6693c665146c3100019f0c92bfa34749a76b3ca/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d_publisher.hpp#L188).

Even though not a percentage anymore I found some uses of the 0-100 range, i.e.: https://github.com/open-navigation/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d_publisher.cpp#L94.

Maybe @DLu can offer some extra info on why a binary value is suggested in the message doc instead of a percentage?

My current naive vote is that we should keep 0-100 and update docs where needed.

DLu: So, now in 2024, almost exactly four years after I made https://github.com/ros2/common_interfaces/pull/117, I now question what the heck I was doing. Specifically with the value of 1. Most frequently, its 100. 

My full thoughts on the values published in Occupancy Grids can be seen in the [robot_navigation documentation](https://github.com/ros2/common_interfaces/pull/117). To stay consistent with the RViz visualization, I would probably stick with 100.

audrow:
> My full thoughts on the values published in Occupancy Grids can be seen in the [robot_navigation documentation](https://github.com/ros2/common_interfaces/pull/117). To stay consistent with the RViz visualization, I would probably stick with 100.

This is a good justification to me to keep things as they are then. I'm going to close this PR. We can reopen it if we decide to revisit this.

Thanks for the suggestion @marcoag!

PR #674, opened 2024-05-01
Switching to example_interfaces

CursedRock17:
This pull request the `demos` side of [this ros2_documentation issue](https://github.com/ros2/ros2_documentation/issues/3358) in which we switch from `std_msgs` to `example_interfaces` because std_msgs was deprecated about 4 years ago.

PR #746, opened 2025-10-22
Add composition::LifecycleTalker component

### LINKED ISSUE #745: Add Lifecycle node component (for testing)
### Description

Currently there is no example for a lifecycle node that can be used as a component.

This could be a simple talker node, which only publishes messages when active.

### Motivation

Currently there is no example lifecycle node component.
This is not a really critical example for how to use the API, as it is relatively straightforward; however, it is essential for testing features such as lifecycle node components in launch_ros, which currently do not behave correctly when namespaced (ros2/launch_ros#479).

Having access to such a lifecycle node component will allow for testing of related features and prevent regression from happening in the future.

### Design / Implementation Considerations

The node would need to be testable such that an effect can be measured after its transition.

An alternative to a new lifecycle node component is to convert the existing lifecycle node to a node component.
However, this would complicate the basic lifecycle node example unnecessarily.

### Additional Information

Related to ros2/launch_ros#479, ros2/launch_ros#480, ros2/launch_ros#481

I could write this; however, I would be unsure in which package to put it.

SuperJappie08:
## Description

Adds a lifecycle component to `composition` (`composition::LifecycleTalker`)

Fixes #745 

### Is this user-facing behavior change?

Yes, a new component is added.

### Did you use Generative AI?

No.

### Additional Information

Currently I have put it in the `composition` package, since it is more of a note that a lifecycle component can exist (and therefore does not extensively focus on the lifecycle part).

### DIFF
--- a/composition/include/composition/lifecycle_talker_component.hpp
+++ b/composition/include/composition/lifecycle_talker_component.hpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef COMPOSITION__LIFECYCLE_COMPONENT_HPP_
-#define COMPOSITION__LIFECYCLE_COMPONENT_HPP_
+#ifndef COMPOSITION__LIFECYCLE_TALKER_COMPONENT_HPP_
+#define COMPOSITION__LIFECYCLE_TALKER_COMPONENT_HPP_
 
 #include "composition/visibility_control.h"
 
@@ -63,4 +63,4 @@ private:
 
 }  // namespace composition
 
-#endif  // COMPOSITION__LIFECYCLE_COMPONENT_HPP_
+#endif  // COMPOSITION__LIFECYCLE_TALKER_COMPONENT_HPP_
--- a/composition/include/composition/lifecycle_talker_component.hpp
+++ b/composition/include/composition/lifecycle_talker_component.hpp
@@ -22,7 +22,7 @@
 #include "rclcpp_lifecycle/lifecycle_node.hpp"
 #include "rclcpp_lifecycle/lifecycle_publisher.hpp"
 
-#include "std_msgs/msg/string.hpp"
+#include "example_interfaces/msg/string.hpp"
 
 namespace composition
 {
@@ -52,7 +52,7 @@ private:
   // is in.
   // By default, a lifecycle publisher is inactive by creation and has to be
   // activated to publish messages into the ROS world.
-  rclcpp_lifecycle::LifecyclePublisher<std_msgs::msg::String>::SharedPtr pub_;
+  rclcpp_lifecycle::LifecyclePublisher<example_interfaces::msg::String>::SharedPtr pub_;
 
   // We hold an instance of a timer which periodically triggers the publish function.
   // As for the beta version, this is a regular timer. In a future version, a
--- a/composition/CMakeLists.txt
+++ b/composition/CMakeLists.txt
@@ -88,7 +88,7 @@ target_link_libraries(lifecycle_talker_component PUBLIC
   rclcpp::rclcpp
   rclcpp_components::component
   rclcpp_lifecycle::rclcpp_lifecycle
-  ${std_msgs_TARGETS})
+  ${example_interfaces_TARGETS})
 rclcpp_components_register_nodes(lifecycle_talker_component "composition::LifecycleTalker")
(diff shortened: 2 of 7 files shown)


SuperJappie08:
I think this is ready for review.
It could use some more tests, but adding tests would make more sense if #747 is fixed.

SuperJappie08:
This requires additional changes in order to align with #674

EDIT: This has been resolved

composition/include/composition/lifecycle_talker_component.hpp
```diff
@@ -0,0 +1,66 @@
...
+  void publish();
+
+  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn on_configure(
+    const rclcpp_lifecycle::State & state) override;
+
+  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn on_cleanup(
+    const rclcpp_lifecycle::State & state) override;
+
+  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn on_shutdown(
+    const rclcpp_lifecycle::State & state) override;
```
mjcarroll, line 45: Should we exercise the other transitions `on_activate`, etc?

composition/include/composition/lifecycle_talker_component.hpp
```diff
@@ -0,0 +1,66 @@
...
+  // We hold an instance of a lifecycle publisher. This lifecycle publisher
+  // can be activated or deactivated regarding on which state the lifecycle node
+  // is in.
+  // By default, a lifecycle publisher is inactive by creation and has to be
+  // activated to publish messages into the ROS world.
+  rclcpp_lifecycle::LifecyclePublisher<example_interfaces::msg::String>::SharedPtr pub_;
+
+  // We hold an instance of a timer which periodically triggers the publish function.
+  // As for the beta version, this is a regular timer. In a future version, a
+  // lifecycle timer will be created which obeys the same lifecycle management as the
```
mjcarroll, line 59: Would definitely appreciate lifecycle timer contribution

mjcarroll (commented):
LGTM, one comment about extra state transitions.

composition/include/composition/lifecycle_talker_component.hpp
SuperJappie08: I'm unsure if I understand correctly, but do you mean 'Should we add the other transitions?'.

I haven't added them right now, as I intended this to be a minimal example for creating lifecycle components.
It isn't a full copy of `lifecycle`'s `LifecycleTalker` as I wanted to keep this simple for testing (for example in ros2/launch_ros#479), because currently there is no suitable lifecycle component to test with.

I'm open to adding these extra transitions, but we should probably consider the 'difficulty order' or complexity of both lifecycle nodes and node components. (Thinking about 'the (fictitious) ROS 2 textbook chapter ordering'.)

For example, I would imagine most users (or at least beginners) would first learn about lifecycle nodes before node components.
This would make it _safe_ to assume they are familiar with lifecycle nodes before looking at this, which would greatly limit the benefit of these nearly empty transitions while introducing more 'clutter'.


(Having written this, I'm second-guessing the ordering of my ROS 2 journey.)

fujitatomoya (commented):
For a demo package, the experience of running it matters a lot.
someone should be able to launch it and immediately see it working, or at minimum know exactly what commands to run.

Looking at how the existing lifecycle demo handles this, it ships a lifecycle_service_client node that automatically drives the state transitions, plus a launch file that brings everything up together. A new user can just ros2 launch and watch the full lifecycle play out.

For this composition::LifecycleTalker, if you just load the component, it starts in the unconfigured state and does nothing visible. Without either an automated driver (a companion node or launch file that loads the component and then transitions it through configure → activate), or clear documentation explaining the exact commands needed (e.g., ros2 component load ..., then ros2 lifecycle set ... configure, then ros2 lifecycle set ... activate) ???

SuperJappie08:
We could add a launchfile using `launch_ros` `LifecycleNodeComponent` autostart or the instructions or both.

I'm not really a fan of the idea of adding another companion node idea, as the `launch` tooling can perform this as well now.

(To add clarification, there is a with autostarting namespaced lifecycle component nodes and I want to add CI tests for it in `launch_ros`, that's the driving point of adding this demo)

EDIT: did better reading and expanded my response

PR #778, opened 2026-03-11
lifecycle_py: Add `ament_mypy` support

mohit-swaran:
## Description
This PR adds ament_mypy support and resolves type-checking errors in the lifecycle_py package
- Add `test/test_mypy.py`
- Used `Dict[str, Any]` in test_lifecycle.py for the `generate_test_description()` return type
- Resolved all `mypy --strict` errors found locally

Fixes (#768)

### Did you use Generative AI?
No

PR #779, opened 2026-03-17
Add ament_mypy support and strict typing to lifecycle_py

### LINKED ISSUE #768: Add `ament_mypy` to `lifecycle_py`
### Description

Test with `mypy --strict . --ignore-missing-imports --allow-untyped-calls`.
Add test_mypy.py file and resolve all regressions. A copy can be found [here](https://github.com/ament/ament_lint/blob/rolling/ament_mypy/test/test_mypy.py)

### Motivation

Improve static typing throughout the ros2 core.

### Design / Implementation Considerations

_No response_

### Additional Information

_No response_

andrearichichi:
## Description

This PR adds `ament_mypy` support to `lifecycle_py` and updates the package to pass strict mypy checks.

Changes include:

* adding `ament_mypy` as a test dependency
* adding a `test_mypy.py` test
* adding type annotations in `talker.py`
* annotating test and launch functions
* limiting `type: ignore` usage to required cases

Validation:

```bash
python3 -m mypy --strict . --ignore-missing-imports --allow-untyped-calls
ament_mypy
```

## Is this user-facing behavior change?

No.

## Did you use Generative AI?

Yes, partially, with manual review.

Closes #768


lifecycle_py/test/test_mypy.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2019 Canonical, Ltd.
```
fujitatomoya, line 1: this is fragile? if the package directory structure ever changes, or if a subdirectory is renamed, this test will silently skip files or fail for structural reasons rather than type errors. this should be replaced with the standard minimal form. IMO, if main(argv=[]) doesn't discover the right files in the ament test harness, that's an ament_mypy bug, not something to work around in every consumer package.

fujitatomoya (changes requested):

InvincibleRMC:
Is a dupe of #778. Feel free to work on an unclaimed typing issue here. You could also start adding typing hints into the basic ros2/examples repo.

